PR #26, opened 2019-07-18
add roadmap

bryanl:


jbeda:
Finally caught up on this and it LGTM.  One suggestion -- we should identify (through examples and experience) different types of plugins.  At some point some simple Plugins might be UI template only and be much easier to author.  Making the plugin experience scalable is going to be important.  simple things simple and hard things possible and all that.

PR #89, opened 2019-07-29
Fix showing only one container in summary view

### LINKED ISSUE #83: Resource viewer loads forever for sidecar containers
Example config:

```
apiVersion: apps/v1
kind: Deployment 
metadata:
  name: sidecar
spec:
  selector:
    matchLabels:
      app: sidecar
  replicas: 2
  template:
    metadata:
      labels:
        app: sidecar
    spec:
      volumes:
      - name: shared-logs 
        emptyDir: {}
    
      containers:
      - name: app-container
        image: alpine
        command: ["/bin/sh"]
        args: ["-c", "while true; do date >> /var/log/app.txt; sleep 5;done"]
    
        volumeMounts:
        - name: shared-logs
          mountPath: /var/log
    
      - name: sidecar-container
        image: nginx:1.16.0
        ports:
          - containerPort: 80
    
        volumeMounts:
        - name: shared-logs
          mountPath: /usr/share/nginx/html
```

GuessWhoSamFoo:
Closes #83

Got bitten by this https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

### DIFF
--- a/internal/modules/overview/printer/pod.go
+++ b/internal/modules/overview/printer/pod.go
@@ -123,8 +123,8 @@ func PodHandler(ctx context.Context, pod *corev1.Pod, opts Options) (component.C
 	o.RegisterItems(conditionDescription)
 
 	var initContainerItems []ItemDescriptor
-	for _, container := range pod.Spec.InitContainers {
-		container := container
+	for i := range pod.Spec.InitContainers {
+		container := pod.Spec.InitContainers[i]
 		cc := NewContainerConfiguration(pod, &container, portForwarder, true, opts)
 		initContainerItems = append(initContainerItems, ItemDescriptor{
 			Width: component.WidthHalf,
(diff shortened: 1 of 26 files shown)


internal/modules/overview/printer/pod.go
```diff
@@ -137,8 +138,9 @@ func PodHandler(ctx context.Context, pod *corev1.Pod, opts Options) (component.C
 
 	var containerItems []ItemDescriptor
 	for _, container := range pod.Spec.Containers {
+		container := container
```
bryanl:
```go
for i := range pod.Spec.Containers {
  container := pod.Spec.Containers[i]
  // ... 
}

File: internal/modules/overview/printer/pod_test.go [hunk 2]
```diff
@@ -15,23 +15,53 @@ import (
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
 	"github.com/vmware/octant/internal/conversion"
 	"github.com/vmware/octant/internal/testutil"
+	"github.com/vmware/octant/pkg/plugin"
+	"github.com/vmware/octant/pkg/store"
 	"github.com/vmware/octant/pkg/view/component"
 )
 
+var (
```
bryanl: These globals in tests was a sloppy pattern I introduced. We should clean these up at every opportunity we get. This should be in a test function somewhere.

GuessWhoSamFoo: Cleaned this up a bit more

PR #99, opened 2019-08-02
add method to use dynamic client instead of informer

bryanl:


internal/dash/dash.go
```diff
@@ -209,8 +209,8 @@ func initObjectStore(ctx context.Context, client cluster.ClientInterface) (store
 	if client == nil {
 		return nil, errors.New("nil cluster client")
 	}
-	
-	appObjectStore, err := objectstore.NewDynamicCache(client, ctx.Done())
+
+	appObjectStore, err := objectstore.NewDynamicCache(ctx, client)
```
wwitzel3: Thanks for fixing that.

Issue #72, opened 2019-07-23
Infinite loading (only header shown), error "crd conversion failed: cannot convert int64 to float64"


hjacobs:
Octant 0.4 does not load for me:

![Screenshot_2019-07-23_17-55-58](https://user-images.githubusercontent.com/510328/61727240-3eebdd00-ad73-11e9-8f40-5b4861d4b596.png)

Console is full of these errors:

```

2019-07-23T17:54:24.950+0200	ERROR	event/generator.go:103	event generator error	{"err": "get CRD from object store: cannot convert int64 to float64", "errVerbose": "cannot convert int64 to float64\nget CRD from object store\ngithub.com/vmware/octant/internal/describer.CustomResourceDefinition\n\t/go/src/github.com/vmware/octant/internal/describer/customresource.go:61\ngithub.com/vmware/octant/internal/describer.(*crdList).Describe\n\t/go/src/github.com/vmware/octant/internal/describer/crd_list.go:57\ngithub.com/vmware/octant/internal/describer.(*CRDSection).Describe\n\t/go/src/github.com/vmware/octant/internal/describer/crd_section.go:62\ngithub.com/vmware/octant/internal/describer.(*Section).Describe\n\t/go/src/github.com/vmware/octant/internal/describer/section.go:35\ngithub.com/vmware/octant/internal/modules/overview.(*realGenerator).Generate\n\t/go/src/github.com/vmware/octant/internal/modules/overview/generator.go:95\ngithub.com/vmware/octant/internal/modules/overview.(*Overview).Content\n\t/go/src/github.com/vmware/octant/internal/modules/overview/overview.go:225\ngithub.com/vmware/octant/internal/event.(*ContentGenerator).generateContent\n\t/go/src/github.com/vmware/octant/internal/event/content_generator.go:75\ngithub.com/vmware/octant/internal/event.(*ContentGenerator).Event\n\t/go/src/github.com/vmware/octant/internal/event/content_generator.go:71\ngithub.com/vmware/octant/internal/event.runGenerator\n\t/go/src/github.com/vmware/octant/internal/event/generator.go:89\ngithub.com/vmware/octant/internal/event.Stream.func1\n\t/go/src/github.com/vmware/octant/internal/event/generator.go:54\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1337"}
github.com/vmware/octant/internal/event.runGenerator
	/go/src/github.com/vmware/octant/internal/event/generator.go:103
[...]

PR #124, opened 2019-08-07
filter crd list of nil entries

wwitzel3:
This prevents octant from crashing when it encounters CRD entires it cannot convert.

related #72 


wwitzel3:
That's what I figured and I figured it was better than looping through the entire list just for the nil cases that may be in there.

PR #126, opened 2019-08-07
Add scoop/choco install and remove version number from install instructions

GuessWhoSamFoo:


wwitzel3:
Is there some alias we can use for the `README.md` so that is is just `latest` instead of needing to bump the version in the file every time?

PR #133, opened 2019-08-08
Update README to clarify installation options for different OSes

nickgerace:
Based off a Twitter exchange, I thought it might be a good idea to clarify installation options. I think the original implementation is good, but this might provide more clarity. 

Reference: https://twitter.com/stretchcloud/status/1159152340995325957?s=20

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,13 @@ Octant is a tool for developers to understand how applications run on a Kubernet
 ## Installation
 
 <!-- TODO: brew and choco install -->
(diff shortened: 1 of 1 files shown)


nickgerace:
I see that this is similar to this PR: https://github.com/vmware/octant/pull/126

Perhaps, there could be a fusion of the two, or their implementation over mine? I apologize for the similar PR!

nickgerace:
Awesome. Thank you!

PR #157, opened 2019-08-12
specify env option for fixing the port

sensay-nelson:


README.md
```diff
@@ -77,6 +77,9 @@ Start running Octant:
 
 `$ octant`
 
+Or to run it on a specific port (8900 in this example):
+`$ OCTANT_LISTENER_ADDR=127.0.0.1:8900 octant`
+
 Octant should immediately launch your default web browser to an available high level port on 127.0.0.1. (i.e. http://127.0.0.1:51234)
```
mikeroySoft: Nice addition, I would recommend this go under line 81 if we want to put this detail up front.
This way we:
- show the octant command
- Set the expectation with what that command does
- then provide variation _(i.e. this PR)_
- follow with the 'for configuring... refer to the docs'

sensay-nelson: Thank you, good idea. I have made the recommended change and also updated the alt host to 0.0.0.0 (as it's common when running in docker).

PR #159, opened 2019-08-13
fix kubeConfig test map cmp

wwitzel3:


bryanl:
@wwitzel3 what as the issue here?

wwitzel3:
It was randomly failing because the other of the items in the map when comparing pointers was different

wwitzel3:
Example https://cloud.drone.io/vmware/octant/181

Issue #142, opened 2019-08-08
Supplied initial namespace flag launches into 'default' namespace


moderation:
We run very restricted clusters where developers only have access to their own namespaces. Access to default or system namespaces is not allowed via RBAC. Similarly access to cluster information is all blocked.

When launching with `./octant -n specific-namespace-name` the browser launches to the URL `http://127.0.0.1:<port>/#/content/overview/namespace/default/?`. The expected behavior would be to launch to `http://127.0.0.1:<port>/#/content/overview/namespace/specific-namespace-name?`

Related: https://github.com/vmware/octant/issues/33, https://github.com/vmware/octant/issues/141

wwitzel3:
Thank you for reporting this issue. We have merged a fix in to master (#165) that will be available in the next binary release ~ August 20th.

moderation:
Fantastic. Thanks very much.

GuessWhoSamFoo:
Fixed via 3ebca1c696e3321f256c23e27cc3ded55fad0786

PR #165, opened 2019-08-14
Handle initial namespace routing with resolver

GuessWhoSamFoo:
This PR uses a resolver to check for the initial namespace from the endpoint `/api/v1/namespace`. It will still attempt redirecting to `/content/overview/namespace/default` until the initial namespace is returned.


wwitzel3:
I believe this will also fix #142, Is that right @GuessWhoSamFoo ?

GuessWhoSamFoo:
Yup, this should also respect the initial namespace flag as well

PR #170, opened 2019-08-15
fix switch clusters with crds

bryanl:
The crd list has to be reset and refreshed when changing clusters. If not, the crd list will not be correct on the cluster that was switched to.

### DIFF
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -14,7 +14,6 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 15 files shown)


internal/cluster/cluster.go
```diff
@@ -155,41 +154,12 @@ func (c *Cluster) ResourceExists(gvr schema.GroupVersionResource) bool {
...
 
 func (c *Cluster) Resource(gk schema.GroupKind) (schema.GroupVersionResource, error) {
-	restConfig, err := c.clientConfig.ClientConfig()
+	restMapping, err := c.restMapper.RESTMapping(gk)
 	if err != nil {
-		return schema.GroupVersionResource{}, errors.Wrap(err, "get rest config")
+		return schema.GroupVersionResource{}, err
 	}
 
-	retries := 0
```
wwitzel3: :+1:

Issue #119, opened 2019-08-07
Add loading indicator (or spinner) while loading resources


alexbrand:
When loading resources from the backend, the UI shows an empty list/table. This could be confusing as it is not clear whether Octant is loading the list of resources or if there are indeed no resources in the cluster.

Screen capture:

![octant-loading](https://user-images.githubusercontent.com/545723/62626633-cf8af700-b8f5-11e9-8954-02706bf7687e.gif)

wwitzel3:
I think having the default data icon be a spinner until the first response / stream from octant to the UI happens is a good idea. As you mention, this would at least signal that there are no resources of that type.

bryanl:
There are multiple locations that should have spinners:

* in navigation
* described resources (i.e the lists you see in the dashboard)

I want to investigate solving them both. More details as I have them

wwitzel3:
This enhancement will be in our next release.

MarcusNoble:
It looks like this may have been removed / broken in the 0.7.0 release.

PR #176, opened 2019-08-16
119 add loading indicators

bryanl:
* how loading indicators when informer cache has not synced
* use less allocations when transversing sections for content
* add placeholders for component tables

note: this doesn't solve the slow loading of views in the web viewer. We'll have to solve that another way.

#119

### DIFF
--- a/internal/describer/customresource.go
+++ b/internal/describer/customresource.go
@@ -18,6 +18,7 @@ import (
 	"github.com/vmware/octant/pkg/store"
 )
 
+// TODO: delete me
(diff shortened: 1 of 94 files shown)


internal/describer/customresource.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/vmware/octant/pkg/store"
 )
 
+// TODO: delete me
```
wwitzel3: Is there an issue for this?

bryanl: ha! there should be. let me go do that now.

Issue #162, opened 2019-08-13
pod logs: can't scroll up after scrolling to bottom


skriss:
When viewing pod logs, if I scroll all the way to the bottom of the logs, the UI doesn't then let me scroll back up towards the top. If I only scroll partially down, I can scroll back up just fine.

I'm on macOS mojave and using chrome.

wwitzel3:
I wasn't able to replicate this with FF or Chrome under Ubuntu are you able to create a gif of the problem? Also would you mind checking the browser console for any errors and pasting those as well? Thank you.

skriss:
Seems like it only occurs if I first navigate to "Resource Viewer" for a pod, then to logs after that. I'll get a gif recorded

wwitzel3:
@nfarruggiagl can you take a look at this and see if you can replicate and fix?

nfarruggiagl:
Sure @wwitzel3 I will take a look!

wwitzel3:
Fixed #189

PR #189, opened 2019-08-19
Fix scroll issue in Logs Container & added unit test for it

nfarruggiagl:
Fix to address #162 

- LogsComponent: the component will now check for changes in `containerLogs` so it knows when a new log is added. 

If the user is not at the bottom of the list, the item will be added to the bottom of it but the scroll position will not change. 
If the user is at the bottom of the list, and a new item is added, the scroll position will remain at the bottom so the user will see all the latests items.
The user should be able to scroll whenever is needed.

- Unit Test added for this functionality.

### DIFF
--- a/changelogs/unreleased/162-nfarruggiagl
+++ b/changelogs/unreleased/162-nfarruggiagl
@@ -1,0 +1,1 @@
+Fix scroll issue in Logs Container & added unit test for it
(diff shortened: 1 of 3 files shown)


GuessWhoSamFoo:
Can you add a changelog entry for this as well?

Issue #63, opened 2019-07-22
Succeeded pods should link from Jobs


wwitzel3:
@GuessWhoSamFoo commented on [Fri Apr 12 2019](https://github.com/heptio/developer-dash/issues/719)

![image](https://user-images.githubusercontent.com/10288252/56055136-98248900-5d0d-11e9-9d16-41ef15368554.png)

Although succeeded pods may be skipped in overview, they should be linked from job especially for logging purposes.

PR #190, opened 2019-08-19
Show succeeded pods under job view only

GuessWhoSamFoo:
#63 

Pushing code for this early because it requires input. Is it correct to implicitly have pods available in another view? Perhaps this should be configurable across all views


wwitzel3:
I think this is a perfectly reasonable approach. @bryanl do you have an opinion here?

PR #191, opened 2019-08-19
default port for Octant is now 7777

### LINKED ISSUE #185: Proposal: create a well defined port for the octant API
Problem:

Octant starts its API on a random port on 127.0.0.1. This avoids potential conflicts. Giving Octant a well defined port number by default means that you could know what the port number is and restarts of the application wouldn't require the browser URL to change. I'm going to suggest two numbers for default ports: 

* 102016 - date Heptio was founded
* 7777 - Kubernetes people seem to like 7s

wwitzel3:
fixes #185 

### DIFF
--- a/internal/dash/dash.go
+++ b/internal/dash/dash.go
@@ -159,7 +159,8 @@ func Run(ctx context.Context, logger log.Logger, shutdownCh chan bool, options O
 
 	listener, err := buildListener()
 	if err != nil {
-		return errors.Wrap(err, "failed to create net listener")
+		err = errors.Wrap(err, "failed to create net listener")
+		return errors.Wrap(err, "use OCTANT_LISTENER_ADDR to set host:port")
 	}
 
 	// Initialize the API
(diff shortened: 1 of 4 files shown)


wwitzel3:
fixes #185

GuessWhoSamFoo:
What should the new behavior be when the port is occupied?

wwitzel3:
We can exit with an error or continue with a warning. I think we should exit with an error if the requested listening port is not available. We can direct the uses to the OCTANT_LISTENER_ADDR setting.

internal/dash/dash.go
```diff
@@ -282,7 +283,12 @@ func initModuleManager(options *moduleOptions) (*module.Manager, error) {
 
 func buildListener() (net.Listener, error) {
 	listenerAddr := api.ListenerAddr()
-	return net.Listen("tcp", listenerAddr)
+	conn, err := net.DialTimeout("tcp", listenerAddr, time.Second*5)
```
bryanl: Couldn’t the timeout be < than 500ms? 5 seconds would slow dow octant’s start up

wwitzel3: yeah, I'll make it 500ms that should be plenty of time

PR #220, opened 2019-08-27
Adds three new resource examples

chelnak:
This PR adds the following new resource examples:

* Config map
* Generic Secret
* Basic multi host ingress

It looks like role and rolebinding have already been created so they can be checked off the list from op.

### DIFF
--- a/examples/resources/generic-secret.yaml
+++ b/examples/resources/generic-secret.yaml
@@ -1,0 +1,10 @@
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: example-secret
+type: generic
+data:
+  API_KEY: bm90aGluZ3Rvc2VlaGVyZQ==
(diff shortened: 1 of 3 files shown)


wsandin:
Hey guys, sorry something urgent come up. Thanks for taking care!

examples/resources/generic-secret.yaml
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2019 VMware, Inc. All Rights Reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+apiVersion: v1
+kind: Secret
+metadata:
+  name: example-secret
+type: generic
+data:
+  API_KEY: bm90aGluZ3Rvc2VlaGVyZQ==
```
rajibmitra: seems like its already base64 encoded ?

chelnak: It just serve as an example.

Decoded it is ‘nothingtoseehere’

chelnak: It just serve as an example.

Decoded it is ‘nothingtoseehere’

PR #226, opened 2019-08-28
fix: make button to stop port forward less ambiguous

theneva:
Playing around with the dashboard, I set up a port forwarding to a pod, then went somewhere else on the dashboard, and came back to this:

image

Without having just seen the button pop up, it was not clear to me what "REMOVE" meant. I still don't have a full overview of what the dashboard can and can't do, so I was pretty afraid to click it.

"Stop port forward" is specific, and matches the wording of "Start port forward" on the button button clicked to add the port forwarding.


theneva:
Aha! I ran the Karma tests, but not the Cypress suite. Thanks :smile:

I've amended my commit to update the Cypress test and include a changelog entry.

PR #243, opened 2019-09-05
add internal error type and error store

### LINKED ISSUE #229: Create store for internal errors.
Currently we log errors and continue. While working to implement #43 which streams errors to a statusview pane in octant, I discovered that things like closing and re-opening the browser window would result in the loss of the error messages being displayed, since that is the nature of streaming the messages.

Octant needs an error store that can hold errors that can be retrieved later. This store can be capped at a reasonable and useful number, maybe the last 100 errors. This will allow other clients to get the list of recent errors even if the client is not listening when the error is originally published on the stream.

The store will implement a simple List/Get set of actions. This will allow a newly connecting client to request the most recent list and allow clients to Get details of errors.

wwitzel3:
closes #229 


GuessWhoSamFoo:
Should the error store have a delete method? It may be jumping ahead to status view work

PR #250, opened 2019-09-10
Remove informer for CR when it no longer exists

bryanl:
When a custom resource no longer exists, remove the informer that pointed to the custom resource.

This change also consolidates operations around determining which custom resources belong to a custom resource definition.


wwitzel3:
Can you add a changelog entry for this.

Issue #235, opened 2019-08-30
Add a HACKING.md


wwitzel3:
This should contain a 1,2,3 breakdown of how to get a dev environment for octant up and running and steps to verify everything is installed correctly and ready to begin development work.

PR #252, opened 2019-09-10
Add instructions to generate web assets

dotNomad:


wwitzel3:
We want to have all the steps to quickly get up and running with contributing in a hacking.md see, #235 

If you could update this PR to do that, that would be very helpful.

PR #271, opened 2019-09-15
use full width if only one container

rajatjindal:
This should fix https://github.com/vmware/octant/issues/261.


internal/modules/overview/printer/pod_template.go
```diff
@@ -104,14 +104,19 @@ func podTemplateContainers(fl *flexlayout.FlexLayout, options podTemplateLayoutO
 
 	containerSection := fl.AddSection()
 
+	width := component.WidthHalf
```
bryanl: This change will still break if you have 3 containers. Instead, if the length of containers is odd, set the width of the last container to full.

rajatjindal: thanks Bryan for the review. i believe its fixed now

PR #277, opened 2019-09-16
Reduce delay caused by retries in dynamic cache get

bryanl:
The delays incurred by retries in dynamic cache get were superfluous so they are removed. Instead, rely on the caller to handle the error received by the action.

### DIFF
--- a/internal/api/context_manager.go
+++ b/internal/api/context_manager.go
@@ -6,7 +6,9 @@
 package api
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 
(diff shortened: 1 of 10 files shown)


internal/api/context_manager.go
```diff
@@ -91,6 +93,8 @@ func (c *ContextManager) Start(ctx context.Context, state octant.State, s Octant
 }
 
 func (c *ContextManager) runUpdate(state octant.State, s OctantClient) PollerFunc {
+	var previous []byte
```
wwitzel3: nice

PR #284, opened 2019-09-17
Set correct content path on websocket reconnect

bryanl:
When the websocket client reconnects, ensure the content path is updated.

This change also:

* removes the `content` prefix from paths
* combines query params, content path, and content updates in a single update.


wwitzel3:
This looks good, just need to address the failures.

bryanl:
@wwitzel3 it's not good. Digging into the nuances

internal/api/filter_manager.go
```diff
@@ -60,20 +61,27 @@ func (fm *FilterManager) Handlers() []octant.ClientRequestHandler {
...
 // ClearFilters clears all filters.
 func (fm *FilterManager) ClearFilters(state octant.State, payload action.Payload) error {
 	state.SetFilters([]octant.Filter{})
+	message := "Cleared filters"
+	state.SendAlert(action.CreateAlert(action.AlertTypeInfo, message, action.DefaultAlertExpiration))
 	return nil
 }
 
 // RemoveFilters removes a filter.
 func (fm *FilterManager) RemoveFilters(state octant.State, payload action.Payload) error {
```
wwitzel3: `RemoveFilter`

internal/api/navigation_manager.go
```diff
@@ -140,7 +139,7 @@ func NavigationGenerator(ctx context.Context, state octant.State, config Navigat
 	for i := range modules {
 		m := modules[i]
 		g.Go(func() error {
-			contentPath := path.Join("/content", m.ContentPath())
+			contentPath := m.ContentPath()
```
wwitzel3: 👍

internal/api/poller.go
```diff
@@ -36,56 +40,138 @@ func (a SingleRunPoller) Run(ctx context.Context, ch <-chan struct{}, action Pol
...
+				if j.ctx.Err() == nil {
+					<-time.After(resetDuration)
+					pollerQueue <- jt.create()
+				}
 			}
-			mu.Unlock()
+		}
+	}
+
+	for i := 0; i < 2; i++ {
```
wwitzel3: make this a const

internal/api/poller.go
```diff
@@ -36,56 +40,138 @@ func (a SingleRunPoller) Run(ctx context.Context, ch <-chan struct{}, action Pol
...
+		for range ch {
+			// cancel all existing jobs
+			jt.clear()
+			pollerQueue <- jt.create()
 		}
 	}()
 
-	done := false
-	for !done {
+	pollerQueue <- jt.create()
```
wwitzel3: I'm confused by this call here. I don't understand why octant needs it if the go func above is clearing and creating job trackers.

internal/api/poller.go
```diff
@@ -36,56 +40,138 @@ func (a SingleRunPoller) Run(ctx context.Context, ch <-chan struct{}, action Pol
...
-				if actionContext.Err() == nil {
-					dur := resetDuration
-					if rerun {
-						dur = 0
-					}
-					timer.Reset(dur)
-				}
-				cancel()
-			}()
+		case <-j.ctx.Done():
```
wwitzel3: do we want to reduce these to a single case?

internal/api/poller.go
bryanl: This is seeding the first job. After a job runs, it queues up the next run.

internal/api/poller.go
bryanl: We need both cases. This particular case means the context for the job was canceled. Nothing else should happen here. The second case will fire if the job was ran.

PR #289, opened 2019-09-18
Replace git.apache.org/trift.git with github.com/apache/thrift

ryanbrainard:
When trying to build this project, I ran into problems with `make go-install` because one of the dependencies (Thrift) uses `git.apache.org`, which is [no longer online](https://status.apache.org/incidents/505pzrgbw9jh) and apparently [not the official repo](https://thrift.apache.org/developers#source-control).

The dependency comes from OpenCensus, and they updated it in https://github.com/census-instrumentation/opencensus-go/commit/3b8e2721f2c3c01fa1bf4a2e455874e7b8319cd7#diff-37aff102a57d3d7b797f152915a6dc16 and released in [v0.19.2](https://github.com/census-instrumentation/opencensus-go/releases/tag/v0.19.2); however, when trying to upgrade, I ran into [this](https://github.com/golang/go/issues/26904):

```shell
$ go get go.opencensus.io@v0.19.2
go: github.com/apache/thrift@v0.12.0 used for two different module paths (git.apache.org/thrift.git and github.com/apache/thrift)
```

I had some luck making it work by just manually find/replacing all references to the repo, but it felt a little hacky and still required this `replace` line you see here. I have a feeling that that was happening because `go get` was still trying to pull from master of this repo. I'm still somewhat new to `go mod`, so I'm open to other suggestions.

So, to keep things simple, I just made the `replace` change in this PR, and then we can probably follow up with an upgrade of OpenCensus.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,3 +61,5 @@ require (
 )
 
 replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+
+replace git.apache.org/thrift.git => github.com/apache/thrift v0.12.0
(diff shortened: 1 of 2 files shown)


bryanl:
I think we should upgrade opencensus to whatever the latest is.

ryanbrainard:
> I think we should upgrade opencensus to whatever the latest is.

I was just trying to make the change as minimal as possible, but I can try that. I'm jumping through some CLA hoops at the moment, so if someone else would like to make that change, that's fine with me.

bryanl:
Fixing in #299

ryanbrainard:
Thanks, I just confirmed #299 fixes the issue.

PR #311, opened 2019-09-23
Add hpa printer

GuessWhoSamFoo:
- [x] Targets for list handler
- [x] Displaying metrics
- [x] Show in resource viewer
- [x] testutil.CreateHPA helper
- [x] Changelog
- [x] component tests
- [x] example resource


internal/link/link.go
```diff
@@ -25,6 +26,7 @@ type Interface interface {
 	ForObjectWithQuery(object runtime.Object, text string, query url.Values) (*component.Link, error)
 	ForGVK(namespace, apiVersion, kind, name, text string) (*component.Link, error)
 	ForOwner(parent runtime.Object, controllerRef *metav1.OwnerReference) (*component.Link, error)
+	ForScaleTarget(parent runtime.Object, scaleTarget *v2beta2.CrossVersionObjectReference) (*component.Link, error)
```
bryanl: This is not general purpose enough to include in this interface. Link generation for scale targets should happen elsewhere.

GuessWhoSamFoo:
A few notes:
 - There is no support for external resources yet because it involves setting up a custom metrics server on a dev environment
 - Octant displays hpa as autoscaling/v1 and adding information from v2beta2 if available. kubectl allows specifying the version explicitly whereas octant cannot
 - Tests are still incomplete but can cover more code paths once review changes are made

GuessWhoSamFoo:
- Made targets available in hpa view
 - Events were not visible despite having events enabled because event object references used autoscaling/v2beta2
 - Changed to scale target and it should handle cases when the scale target does not exist

PR #313, opened 2019-09-23
create a ClusterClientManager

wwitzel3:
First step in breaking out the global state of Octant

### DIFF
--- a/internal/config/dash.go
+++ b/internal/config/dash.go
@@ -74,6 +74,8 @@ type Dash interface {
 
 	ClusterClient() cluster.ClientInterface
 
+	ClusterClientManager() client.ClusterClientManager
+
 	CRDWatcher() CRDWatcher
 
 	ObjectStore() store.Store
@@ -95,8 +97,6 @@ type Dash interface {
 	Validate() error
 
 	ModuleManager() module.ManagerInterface
-
-	// ClusterClientManager()
 }
 
 // Live is a live version of dash config.
@@ -110,7 +110,7 @@ type Live struct {
 	pluginManager        plugin.ManagerInterface
 	portForwarder        portforward.PortForwarder
 	kubeConfigPath       string
-	currentContextName   string // remove .. replace with CurrentContext() method ..
+	currentContextName   string
 	restConfigOptions    cluster.RESTConfigOptions
 }
 
@@ -118,7 +118,7 @@ var _ Dash = (*Live)(nil)
 
 // NewLiveConfig creates an instance of Live.
 func NewLiveConfig(
-	clusterClientManger client.ClusterClientManager,
+	clusterClientManager client.ClusterClientManager,
 	crdWatcher CRDWatcher,
 	kubeConfigPath string,
 	logger log.Logger,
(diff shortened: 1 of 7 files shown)


internal/config/dash.go
```diff
@@ -94,49 +97,59 @@ type Dash interface {
 	Validate() error
 
 	ModuleManager() module.ManagerInterface
+
+	// ClusterClientManager()
```
bryanl: If this is going to be implemented in a future change, create a todo?

internal/config/dash.go
```diff
@@ -94,49 +97,59 @@ type Dash interface {
...
+		kubeConfigPath:       kubeConfigPath,
+		logger:               logger,
+		moduleManager:        moduleManager,
+		objectStore:          objectStore,
+		pluginManager:        pluginManager,
+		portForwarder:        portForwarder,
+		currentContextName:   currentContextName, // remove
+		restConfigOptions:    restConfigOptions,
+	}
+	clusterClient, err := l.clusterClientManager.Get(context.Background(), currentContextName)
```
bryanl: This context should come from somewhere else. Don't use a `Background` here.

internal/config/dash.go
wwitzel3: It is, it is, just a left over comment. Line 22 is identical (and uncommented)

internal/config/dash.go
wwitzel3: Made it a TODO context for now, will get a proper context plumbed down to it.

internal/config/dash.go
```diff
@@ -94,49 +97,59 @@ type Dash interface {
...
+	clusterClient        cluster.ClientInterface // clusterClientManager
+	clusterClientManager client.ClusterClientManager
+	crdWatcher           CRDWatcher
+	logger               log.Logger
+	moduleManager        module.ManagerInterface
+	objectStore          store.Store
+	pluginManager        plugin.ManagerInterface
+	portForwarder        portforward.PortForwarder
+	kubeConfigPath       string
+	currentContextName   string // remove .. replace with CurrentContext() method ..
```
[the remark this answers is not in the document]
wwitzel3: I can do it in the this PR, but my plan was to do this in a separate PR. I can just continue the work as a new commit on this same PR.

internal/config/dash.go
```diff
@@ -186,23 +209,18 @@ func (l *Live) PortForwarder() portforward.PortForwarder {
 
 // UseContext switches context name. This process should have synchronously.
 func (l *Live) UseContext(ctx context.Context, contextName string) error {
-	client, err := cluster.FromKubeConfig(ctx, l.kubeConfigPath, contextName, l.restConfigOptions)
-	if err != nil {
-		return err
-	}
-
-	l.ClusterClient().Close()
```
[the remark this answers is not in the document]
wwitzel3: Yeah, I can address this in this PR. `newCluster` I think will become `Open` and I'll have the `FromKubeConfig` function give back a clientConfig / restConfig .. the cluster client manager can then store those and `Open` the a client interface on demand. As for closing, we could add a timeout mechanism that closes clients that have gone unused for a period of time.

PR #317, opened 2019-09-25
add relative/absolute handling to LinkComponent

### LINKED ISSUE #315: Problem adding an external link to SummarySection
I'd like to add a link to an external site into a row on the status panel.

Trying to use `NewLink`:

```go
{Header: "Title", Content: component.NewLink("", "Example", "https://example.com")}
```

Results in a relative link, not an absolute one, like so: http://127.0.0.1:7777/#/content/overview/namespace/default/workloads/deployments/<name>/https://example.com

Trying to use `NewMarkdownText` instead:

```go
{Header: "Title", Content: component.NewMarkdownText("[Example](https://example.com)")}
```

The absolute link works here, but results in wrapping the link in a paragraph, which adds lots of additional padding to the top of the text when rendered in the table.

It would be nice to have a way of adding nicely formatted external links to a panel, by fixing either of the above behaviours or by adding a new `NewExternalLink`.

wwitzel3:

fixes #315

### DIFF
--- a/web/src/app/modules/overview/components/link/link.component.html
+++ b/web/src/app/modules/overview/components/link/link.component.html
@@ -1,1 +1,7 @@
-<a [routerLink]="[ref]">{{ value }}</a>
+<ng-template [ngIf]="isAbsolute" [ngIfElse]="relative">
+    <a href="{{ ref }}">{{ value }}</a>
+</ng-template>
+
+<ng-template #relative>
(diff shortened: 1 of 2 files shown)


wwitzel3:
This is how I did it in my first pass and then for some reason I switched it to it's own component. I can not at this moment explain why.

PR #328, opened 2019-10-04
Added notifier warning when user adds non-existent namespace

GuessWhoSamFoo:
This handles updating the warning as namespaces are created and deleted.

image


wwitzel3:
When I first start Octant, I get the warning when visiting `http://localhost:7777/`. Is there away we can suppress that warning until after the initial load of `default` or what ever the INITIAL_NAMESPACE is set to.

GuessWhoSamFoo:
Also includes a fix for #142

PR #336, opened 2019-10-09
Address frontend memory leak

GuessWhoSamFoo:
Goal of this PR is to ensure JS heap size does not keep increasing if octant is left idle in a browser. 

currently:
image

0.7
image

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ jobs:
   build:
     machine:
       enabled: true
-      docker_layer_caching: true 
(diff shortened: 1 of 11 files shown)


GuessWhoSamFoo:
@wwitzel3 The dev environment feels a lot better and less prone to crashing. 

There is a post (https://stackoverflow.com/questions/14034107/does-javascript-setinterval-method-cause-memory-leak) suggesting `setInterval` is what causes heap usage to appear growing but it might actually by fine.

image

PR #342, opened 2019-10-11
reduce content refresh delay to 1s

bryanl:
This change will reduce the lag in dashboard view for a (tiny) increase in CPU.

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -8,4 +8,4 @@ package event
 import "time"
 
 // DefaultScheduleDelay is the default schedule delay
-const DefaultScheduleDelay = 5 * time.Second
+const DefaultScheduleDelay = 1 * time.Second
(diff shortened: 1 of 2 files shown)


GuessWhoSamFoo:
Can't reproduce it consistently, but there might be a case where switching namesapces via dropdown doesn't switch

bryanl:
> Can't reproduce it consistently, but there might be a case where switching namesapces via dropdown doesn't switch

Is this why cypress is failing?

GuessWhoSamFoo:
Adding below to https://github.com/vmware-tanzu/octant/blob/master/web/cypress/integration/app_spec.js#L31 should fix the breaking test

```
        cy
            .location('hash')
            .should('include', '/' + namespace)
```

GuessWhoSamFoo (approved):
Re-ran the test and it passed :poop:

PR #343, opened 2019-10-11
add nightly build to drone

wwitzel3:


.drone.yml
```diff
@@ -43,7 +40,47 @@ steps:
...
+        from_secret: google_application_json
+    trigger:
+      event:
+      - cron
+      cron:
+      - nightly
+    image: goreleaser/goreleaser:v0.119-cgo
+    commands:
+      - echo $GOOGLE_APPLICATION_JSON > gs.json
+      - GOOGLE_APPLICATION_CREDENTIALS=gs.json goreleaser -f .goreleaser-nightly.yml --rm-dist --skip-validate
```
bryanl: I can't see the credentials in drone, but there would be two things to check:
- Are the credentials being retrieved correctly?
- Should you use a full path for `GOOGLE_APPLICATION_CREDENTIALS`?

wwitzel3: I moved everything in to a bash script to give me better control over it. So I'll see if that helps after this next run.

Issue #205, opened 2019-08-23
Allow all environment variables to also be passed as flags.


wwitzel3:
Using https://github.com/spf13/viper

Octant can call `SetEnvPrefix` and `AutomaticEnv` to automatically map `viper.Get` calls to ENV variables that are upper case versions of our flag names.

This change has the added benefit of reducing the configuration surface area in the code. We can use `viper.Get` for all values - flags and ENV variables.

related #204

PR #351, opened 2019-10-16
Add option to not open browser

johscheuer:
This options allows to specify wether the a browser should be opened or not in stead of hardcoding it to true. This is useful for environments where no web browser is installed (and prevent an ugly error message).

### DIFF
--- a/internal/commands/dash.go
+++ b/internal/commands/dash.go
@@ -32,6 +32,7 @@ func newOctantCmd(version string) *cobra.Command {
 	var initialContext string
 	var klogVerbosity int
 	var clientQPS float32
+	var openBrowser bool
 	var clientBurst int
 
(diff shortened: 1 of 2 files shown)


wwitzel3:
Thank you for the PR. We have an environment variable that can be set, `OCTANT_DISABLE_OPEN_BROWSER` (See: https://github.com/vmware-tanzu/octant/blob/master/docs/getting-started.md#environment-variables). We plan to have this option as a flag as well, as part of #205 . If you are interested in working on that issue?

johscheuer:
Ah thanks ! I didn't saw this and directly looked into the code (was confused that the value was overwritten at another place).

I will take a look and if I find some time I will provide a patch doesn't seem to be that complicated.

Thanks!

PR #386, opened 2019-10-29
fix js mime type for Windows

### LINKED ISSUE #355: Blank page in browser (0.8.0, Windows)
I've downloaded Octant 0.8.0 and it does only show a blank page in all my browsers (edge dev, chrome, firefox)

From the log it seems as if the Content-Type for JavaScript-files is wrong. Here's an excerpt:

Loading module from “http://127.0.0.1:7777/main-es2015.f3cf8b8bca1b3debed1e.js” was blocked because of a disallowed MIME type (“text/plain”).

I'm using Octant on Windows 10.

wwitzel3:
Fixes #355 

### DIFF
--- a/changelogs/unreleased/355-wwitzel3
+++ b/changelogs/unreleased/355-wwitzel3
@@ -1,0 +1,1 @@
+fix javascript mime type for Windows
--- a/web/web.go
+++ b/web/web.go
@@ -1,9 +1,10 @@
 package web
 
 import (
+	"mime"
 	"net/http"
 
-	"github.com/GeertJohan/go.rice"
+	rice "github.com/GeertJohan/go.rice"
 )
 
 //go:generate rice embed-go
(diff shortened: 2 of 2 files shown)


GuessWhoSamFoo (approved):
Can't seem to reproduce this on my windows setup but lgtm

wwitzel3:
I could not reproduce either, but if you look at #355 you'll see the reporter was able to build my patch and test that is fixed the issue for them.

maehld:
I hope it is ok to leave a comment here? Regarding the reproducibility. I checked the golang's mime code and found the code which gets the correct mime type from the registry. The key says "text/plain" on my system. I don't know why but I certainly didn't insert the key myself :-)

Here's a screenshot:

image

Maybe one could open a report with the golang team? They could just ignore the registry for well known content types or is that a bad thing?

wwitzel3:
This is really good info, thank you @maehld

Issue #324, opened 2019-10-02
Expose launch command in public api


GarySmith:
In airshipui we are trying to distribute our ui as a single go executable. But before launching directly into octant there is some pre-launch logic that needs to be executed. So the plan is to write a small amount of wrapper code that imports octant and launches it octant by calling internal/commands/dash/newOctantCmd. But since this is in an internal package, it is not callable by another package. 
If this were exposed in the public api (pkg/...), then it would enable this usage.

wwitzel3:
I think this should be possible, we'll need to consider the public API of an OctantCmd and look at what we currently are exposing as part of that.

wwitzel3:
We will be tracking and doing this work on issue #204.

PR #396, opened 2019-10-31
allow 3rd party code to easily embed Octant

### LINKED ISSUE #204: Allow the dashboard to be embedded in another program
I started to check if it was possible to embed Octant inside another Go application but right now a large portion of the code sits inside the `internal/` folder which prevents it do be imported.

The goal would be to automatically start/stop Octant when starting something else. It would allow things like an embedded Octant inside a Grafana plugin for instance.

I think moving `internal/dash` to `pkg/dash` would be enough by looking at this: https://github.com/vmware/octant/blob/d52b4b0286f865a3d2ad7d3ffb8d6e8bc1eb48ea/internal/commands/dash.go#L68-L92

Would you be open to that change? I'm happy to drive that change and the testing related to it 🙂

wwitzel3:
closes #204 #324 


GuessWhoSamFoo:
@GarySmith Was this what you had in mind?

wwitzel3:
@GarySmith @ledor473 does this enable what you all need?

ledor473:
@wwitzel3 I think it does, I would have to check if it's importable now and if I can run it, but from a quick look at the changes it appears fine

GuessWhoSamFoo (approved):
Let's make this available for now and we can always revert before the next release if it is not useful

ledor473:
@wwitzel3 @GuessWhoSamFoo sorry for the delay, while I don't have the time to fully test it, the import went well. I just have to make our logger match your interface and once that's done it should work.

PR #408, opened 2019-11-07
use build.go instead of a Makefile

### LINKED ISSUE #256: Windows hacking instructions.
Extend our HACKING.md to include specific instructions for Windows.

We may want to also consider adding a Windows specific Makefile?

wwitzel3:
fixes #256 


Makefile
```diff
@@ -1,109 +1,54 @@
 # Copyright (c) 2019 the Octant contributors. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
```
bryanl: Why do you need a Makefile?

wwitzel3: it's just mocked out while I test things again CI , it will get removed when I go to mark this ready for review

hacks/test-runner.sh
```diff
@@ -27,14 +27,14 @@ should_build() {
 }
 
 if [[ $TRAVIS_BRANCH == 'master' ]]; then
```
GuessWhoSamFoo: This file can be deleted. We could adapt it for use with drone, but unit tests are running at an acceptable speed to not need this in the foreseeable releases.

GuessWhoSamFoo (approved):

PR #418, opened 2019-11-08
[WIP] Terminal / Execute command feature

wwitzel3:


GuessWhoSamFoo:
Punting on having a component link directly to a tab because angular routerLink does not have a way to turn off encoding. @nfarruggiagl and I are going to look into this

GuessWhoSamFoo:
Karma is saying the view config from ngOnInit is underfined in ngAfterViewInit. Looking to see if the afterviewinit lifecycle hook is actually necessary

GuessWhoSamFoo:
Items left todo:

 - [ ] Ability to delete a terminal
 - [ ] Pass karma tests to enable build
 - [ ] Check if tty is checked and only load scrollback

GuessWhoSamFoo:
The container tests will need to have actions added to them

wwitzel3:
Instead of a terminals tab we are going to add new tends to the slider that was merged in #440

GuessWhoSamFoo:
The summary tab currently crashes with the sample plugin installed for some unknown reason

wwitzel3:
- [ ] slider should fill space when resized
- [ ] terminal component should fill space when in a slider that is resized
- [ ] reszie should only be allowed when slider is out
- [ ] size should be preserved after resize on slide-in/slide-out 
- [ ] active tab in tabs component currently uses #anchor - extension tabs need to use something else
- [ ] tabs need to be set as closable
- [ ] closing a tab needs to call the delete action

wwitzel3:
replaced by #487

Issue #404, opened 2019-11-05
Event messages are sorted inconsistently


GuessWhoSamFoo:
![event-sort](https://user-images.githubusercontent.com/10288252/68255951-8aa15a00-ffe3-11e9-959e-d1acf8e992f6.gif)

Emitted events should appear chronologically and should not move around on a table component.

alexbrand:
Was this fixed by #419?

wwitzel3:
Yes.

PR #419, opened 2019-11-09
Sort events in a more stable manner

bryanl:
#404

### DIFF
--- a/build.go
+++ b/build.go
@@ -64,6 +64,8 @@ func main() {
 			test()
 		case "build":
 			build()
+		case "run-dev":
+			runDev()
 		case "go-install":
 			goInstall()
 		case "serve":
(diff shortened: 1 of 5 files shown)


build.go
```diff
@@ -135,6 +137,10 @@ func build() {
 	runCmd("go", nil, "build", "-o", "build/octant", GO_FLAGS, "-v", "./cmd/octant")
 }
 
+func runDev() {
+	runCmd("build/octant", nil)
```
wwitzel3: Can you make this check for the existence of `build/octant` and run `build` if it isn't find?

wwitzel3 (approved):
This works nicely for me, I made a suggestion, but not a blocker

build.go
bryanl: I updated the code and split the commits. Now `newCmd` will check for the existence of the command first.

PR #429, opened 2019-11-16
Do not remove fakes when running "build.go clean"

antoninbas:
This is a follow-up top #428 
Now that fakes are checked-in, running "clean" should not remove them.

Signed-off-by: Antonin Bas <abas@vmware.com>


antoninbas:
My bad, I should have caught this too...

PR #450, opened 2019-11-27
remove in-cluster example and update documentation

wwitzel3:
Flags and variables that were intended to aid development have leaked out to the user level. Octant is an application and is intended as a single client tool and at this time we have no plans to support hosted versions of Octant.

Currently we use the default web browser on the system as our UI client. Octant will eventually be packaged with a client, probably using Electron or similar. This will further abstract away the details of the internals and hopefully remove the confusing messaging around how to run Octant.

There is possibly that in the future, we would add some ability to remotely share/view another Octant, but that would look more like what you see with vscode live share or tmate sharing.


bryanl (approved):

aleveille:
For what it's worth, I've made a ready-to-use published Docker image and Helm chart to deploy to Kubernetes: https://github.com/aleveille/octant-dashboard-turnkey

For anyone looking to deploy to Kubernetes, you can hop over there as I will likely maintain that repo for a couple months/years.

Issue #405, opened 2019-11-06
Custom Resource view doesn't align with kubectl


alanmeadows:
The view octant provides for custom resources is limited in usefulness as it does not take advantage of the rich column data that kubectl can display defined within the CRD.  My expectation is that there would be parity between these two views in terms of a listing of objects.

For instance:

![image](https://user-images.githubusercontent.com/716166/68324446-3dbd9200-007c-11ea-918b-ca021a92daa9.png)

vs:

```
alan@m3-io-demo-2:~/metal3-dev-env$ kubectl get baremetalhost -n metal3
NAME       STATUS   PROVISIONING STATUS   CONSUMER   BMC                         HARDWARE PROFILE   ONLINE   ERROR
master-0   OK       ready                            ipmi://192.168.111.1:6230   unknown            false    
worker-0   OK       ready                            ipmi://192.168.111.1:6231   unknown            true     
alan@m3-io-demo-2:~/metal3-dev-env$ 
```
This would be expected to have been provided by the CRD as printer columns, such as the case for the `baremetalhosts.metal3.io` below -- perhaps defaulting back to the view that is currently available on Octant when these are not defined in a CRD.  However, in this case, the CRD owner has informed us of what is important in terms of columns already.

```
...
  name: baremetalhosts.metal3.io
  resourceVersion: "692"
  selfLink: /apis/apiextensions.k8s.io/v1/customresourcedefinitions/baremetalhosts.metal3.io
  uid: ebf2510d-fcf2-4e8b-a174-98187657989a
spec:
  conversion:
    strategy: None
  group: metal3.io
  names:
    kind: BareMetalHost
    listKind: BareMetalHostList
    plural: baremetalhosts
    shortNames:
    - bmh
    - bmhost
    singular: baremetalhost
  preserveUnknownFields: true
  scope: Namespaced
  versions:
  - additionalPrinterColumns:
    - description: Operational status
      jsonPath: .status.operationalStatus
      name: Status
      type: string
    - description: Provisioning status
      jsonPath: .status.provisioning.state
      name: Provisioning Status
      type: string
    - description: Consumer using this host
      jsonPath: .spec.consumerRef.name
      name: Consumer
      type: string
    - description: Address of management controller
      jsonPath: .spec.bmc.address
      name: BMC
      type: string
    - description: The type of hardware detected
      jsonPath: .status.hardwareProfile
      name: Hardware Profile
      type: string
    - description: Whether the host is online or not
[...]

PR #473, opened 2019-12-12
List additional columns in Custom Resource view

GarySmith:
Custom Resource Definitions may additionalPrinterColumns that are
specific to a give api version.  The CustomResourceHandler and
CustomResourceListHandler were only handling top-level
additionalPrinterColumns that apply to all columns regardless of
version.  This change handles version-specific additionalPrinterColumns.
This change addresses https://github.com/vmware-tanzu/octant/issues/405

### DIFF
--- a/internal/printer/customresource.go
+++ b/internal/printer/customresource.go
@@ -21,6 +21,56 @@ import (
 	"github.com/vmware-tanzu/octant/pkg/view/component"
 )
 
+func getAdditionalPrinterColumnsForList(
+	crd *apiextv1beta1.CustomResourceDefinition,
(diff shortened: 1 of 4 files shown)


GarySmith:
Note that this is failing the unit test due to an unrelated bug in the internal representation of CRDs in octant, discussed here https://github.com/vmware-tanzu/octant/issues/405#issuecomment-564126293 . Wayne Witzel suggested in the octant slack channel that this is being addressed in another PR.

bryanl:
Hello, I'm going to close this PR because I want to fix a bigger problem around CRDs that will make this change redundant. Thanks for pointing this issue out, and we'll have a fix pretty soon.

bryanl:
I revamped the CRD system in #490. Should solve this problem and a couple of others. Please retest when you get a chance.

PR #488, opened 2019-12-17
Terminal feature

### LINKED ISSUE #43: Octant Status View
Surface the status of Octant, starting with logging information to the user in the UI.

### LINKED ISSUE #135: Feature request: Terminal connect to running container
Thanks for this great tool!
Is it possible to add a terminal tab to connect to running container to execute some scripts inside?

wwitzel3:
Adds a terminal component.
Adds the ability to execute a command against a container.
Adds a separate extension view for holding terminals, displayed as tabs at the bottom of Octant.

terminal-exec-demo

fixes #135 
closes #43


wwitzel3:
There are some known issues that we will be creating issues for after this PR is merged:
- xtermjs console error when resizing the slider view.
- remote resize causes interactive shells to freeze (currently commented out).
- after resizing the slider, if you navigate to a new page, the slider size is forgotten.
- current tab in the extension view is forgotten when navigating to a new page.

internal/api/terminal_manager.go
```diff
@@ -0,0 +1,174 @@
...
+
+func (s *terminalStateManager) Start(ctx context.Context, state octant.State, client OctantClient) {
+	ch := make(chan struct{}, 1)
+	defer func() {
+		close(ch)
+	}()
+	s.poller.Run(ctx, ch, s.runUpdate(state, client), event.TerminalStreamDelay)
+}
+
+func (s *terminalStateManager) runUpdate(state octant.State, client OctantClient) PollerFunc {
```
wwitzel3: I think using a single poller to loop over terminals to send the websocket message is not great. I'd rather see this done as an individual websocket endpoint for each terminal.

I imagine this happening in the shape of a websocket broker that can register wss endpoints on demand.

internal/api/terminal_manager.go
```diff
@@ -0,0 +1,174 @@
...
+			data := terminalOutput{Line: line}
+
+			if ok && sendScrollback.(bool) {
+				data.Scrollback = t.Scrollback()
+				msg := t.ExitMessage()
+				if msg != "" {
+					data.Scrollback = append(data.Scrollback, []byte("\n"+msg)...)
+				}
+				s.setSendScrollback(t.ID(), false)
+			}
```
wwitzel3: The logic that decides if to send scrollback vs. line data is a bit hard to follow. There should just be a well defined interface that the TerminalManager calls to get a terminal event and then sends that event over the websocket.

This is leaking way too much detail in to the manager.

bryanl:
Some observations:

* The terminal has about a quarter second delay between typing a character and it showing up on the screen. Tried it on two devices in multiple browsers
* When I run a command, I'd expect the terminal to open as soon as I hit enter. If it didn't open, I'd click the the thing that changed (the title of the terminal), and then maybe I'd figure out that you need to click on the thing on the bottom right. This does not feel intuitive to me.
* Using the terminal creates a large amount of network traffic between the cluster and octant. I'm currently not even using the terminal, but it's sending approximately 10MB/s traffic to the cluster. 
* The terminal bar at the bottom of the screen should not cover up any content. It should not be over the bottom of the page when closed, but rather at the bottom of the page below the navigation and overview content areas. 
* When the terminal exits, it should say that it exited and also change the title to show that it has exited. 
* Multiple Octant browsers windows share a set of tabs. When multiple windows are open at once, the input works, but the output looks like it randomly displays parts of it on all the open open windows.

wwitzel3:
I believe I have addressed the input lag in a commit I pushed up, but I was not experiencing input lag before, so I will need you to test it for me @bryanl.

I have also pushed up a quick adjustment so the terminal / tab show that they have exited. It introduced a new bug that will need to be resolved. Changing the tab name makes angular and clarity redraw the component instead of updating it and the result it you get bumped off the tab.

The bottom bar does not cover up any content for me in Chrome or Firefox. Can you please provide a screenshot and browser details.

wwitzel3:
The multiple tab issue is something we probably want to address on a larger scale. Even without using the terminal, if you run Octant with `go run -race` and open multiple tabs and start clicking around there are 4 data races that are found and the result is odd behavior on the front end in some cases.

I am not dismissing it and will create an issue for it but I don't think that is a blocker here. These issue of a single Octant with multiple tabs also goes away once we make Octant an Electron application. Since there would be essentially the equivalent of an Octant per tab.

wwitzel3:
Had a conversation with @bryanl and he thinks that the switching tabs causing stream jitters is a bad experience. After thinking about it more, I agree. Although are intent is to have Octant be an application our reality is that the browser is currently our UI client and we must try to make that experience good while that's what we have.

He also pointed out a bug in which we do not properly close out the terminals when switching contexts.

wwitzel3:
@GuessWhoSamFoo has discovered that using the v2 version of the stream protocol removes the performance issues we've been seeing.

wwitzel3:
We may not need to pin the stream version. It appears that there is just an implementation problem with our TerminalSizeQueue / TerminalSize handling

pkg/view/component/terminal_test.go
```diff
@@ -0,0 +1,51 @@
...
+		Container: "container-id",
+		Command:   "/bin/bash",
+		UUID:      "0000-0000-0000-0000-0000",
+		Active:    false,
+	}
+	input := NewTerminal("default", "term-test", details)
+	actual, err := json.Marshal(input)
+	assert.NoError(t, err)
+
+	var expected = `
```
bryanl: If you put this in `testdata/whatever`, you could load it from the filesystem and not have to deal with weird formatting in a string.

bryanl:
Also check out the following typescript file as they are not formatted by prettier:

* web/src/app/components/quick-switcher/quick-switcher.component.ts
* web/src/app/models/resource-node.ts
* web/src/app/modules/overview/overview.module.ts
* web/src/app/modules/overview/services/websocket/websocket.service.ts

pkg/view/component/component.go
```diff
@@ -18,10 +18,11 @@ var EmptyContentResponse = ContentResponse{}
 // ContentResponse is a a content response. It contains a
 // title and one or more components.
 type ContentResponse struct {
-	Title      []TitleComponent `json:"title,omitempty"`
-	Components []Component      `json:"viewComponents"`
-	IconName   string           `json:"iconName,omitempty"`
-	IconSource string           `json:"iconSource,omitempty"`
+	Title              []TitleComponent `json:"title,omitempty"`
+	Components         []Component      `json:"viewComponents"`
+	ExtensionComponent Component        `json:"extensionComponent,omitempty"`
```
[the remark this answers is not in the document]
GuessWhoSamFoo: This corresponds to the slider on the bottom (See: https://github.com/vmware-tanzu/octant/pull/488/files#diff-da7ac812519f147d17569f6824b131b0R263). It has all the properties of view components except it does not use hash fragments to track tabs.

wwitzel3:
After pairing on this last night, there is only one issue left with the terminal.

- [ ] When I run a command, I'd expect the terminal to open as soon as I hit enter. If it didn't open, I'd click the the thing that changed (the title of the terminal), and then maybe I'd figure out that you need to click on the thing on the bottom right.

GuessWhoSamFoo:
Bonus feature: what key binding should be mapped to the slider toggle?

GuessWhoSamFoo:
The slide service tracks the active tab throughout overview. There is still some cleaning needed for various classes of closing tabs.

web/src/app/modules/overview/components/tabs/tabs.component.ts
```diff
@@ -61,9 +72,30 @@ export class TabsComponent implements OnChanges, OnInit {
...
         this.setMarker(this.activeTab);
       }
+
+      if (this.extView && this.tabs.length > 0) {
+        this.sliderService.activeTab.subscribe(index => {
+          this.activeTabIndex = index;
+        });
+
+        // Case where new tab is added or initial load
+        if (views.length > preViews.length || this.activeTabIndex === null) {
```
GuessWhoSamFoo: I think this is the problem because it cannot distinguish between closing a tab vs opening one which messes up the index

bryanl (approved):
:shipit:

PR #490, opened 2019-12-17
Handle mutliple CRD versions

bryanl:
CRD api has moved from v1beta1 to v1. Handle both methods.

This change also introduces a new display for CRDs. You can now see some CRD configuration and multiple versions of custom resources.

Screen Shot 2019-12-17 at 2 36 04 PM


wwitzel3 (approved):
Minor comments can merge as-is or fix.

PR #573, opened 2020-01-24
bug fix: don't let latent pod metrics break the workload viewer

### LINKED ISSUE #572: bug: in certain situations, workloads viewer will crash with new pods
```
2020-01-24T08:53:53.528-0800    ERROR   api/content_manager.go:103      generate content        {"client-id": "d5b13978-3ec9-11ea-83bc-acde48001122", "err": "generate content: load cards cards: collect workload cards: load workloads: get metrics for pod 'operatorhub-bb644598f-qvqqw': get pod metrics: get pod metrics: podmetrics.metrics.k8s.io \"default/operatorhub-bb644598f-qvqqw\" not found", "content-path": "workloads/namespace/default"}
        /Users/bryan/Development/projects/octant/internal/api/content_manager.go:103
github.com/vmware-tanzu/octant/internal/api.(*job).run.func1
        /Users/bryan/Development/projects/octant/internal/api/poller.go:111
```

The cause is that the metrics server has yet to generate metrics for the pod. This is not an error and Octant should be more patient

bryanl:
This change returns gracefully if pod metrics are not available for a pod. (i.e., it is new and they don't exist yet)

Fixes: #572

### DIFF
--- a/changelogs/unreleased/572-bryanl
+++ b/changelogs/unreleased/572-bryanl
@@ -1,0 +1,1 @@
+bug fix: don't fail if pod metrics are not available for a pod (they are not available instantly)
--- a/internal/octant/fake/mock_pod_metrics_crud.go
+++ b/internal/octant/fake/mock_pod_metrics_crud.go
@@ -34,12 +34,13 @@ func (m *MockPodMetricsCRUD) EXPECT() *MockPodMetricsCRUDMockRecorder {
 }
 
 // Get mocks base method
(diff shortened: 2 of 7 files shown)


wwitzel3 (approved):

GuessWhoSamFoo:
Screenshot_20200124_131101

Not blocking: each card should be wrapped in a flexlayout using width half so that sizing does not change if a workload is missing metrics.

bryanl:
@GuessWhoSamFoo what we should do is say that metrics aren't yet available and give an excuse why. This solves the immediate issue and I will have a new issue for the follow-up.

PR #574, opened 2020-01-24
Fix typo on GO111MODULE in build script's goInstall()

ilayaperumalg:
Signed-off-by: Ilayaperumal Gopinathan <igopinathan@pivotal.io>


GuessWhoSamFoo:
This went mostly unnoticed as builds used go 1.13 so it was set as auto despite the type

PR #588, opened 2020-01-28
Fixed CRD watcher race condition

GuessWhoSamFoo:
Also increases timeout of cypress commands from 25 to 50 seconds with e2e test

Co-authored-by: bryanl <bryanliles@gmail.com>

### DIFF
--- a/web/cypress/integration/crd.spec.js
+++ b/web/cypress/integration/crd.spec.js
@@ -1,0 +1,37 @@
+describe('CRD', () => {
+    before(() => {
+        cy.visit('/');
+    })
+})
+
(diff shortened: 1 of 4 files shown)


web/cypress/integration/crd.spec.js
```diff
@@ -0,0 +1,37 @@
+describe('CRD', () => {
+    before(() => {
+        cy.visit('/');
+    })
+})
+
+it('loads CRD', () => {
```
bryanl: Qthis test would not have caught the CRD path issue.  CRDs added after Octant booted worked fine.

PR #589, opened 2020-01-28
WIP backoff

wwitzel3:
fixes #579 
Signed-off-by: Wayne Witzel III <wayne@riotousliving.com>

PR #593, opened 2020-01-29
backoff for objectstore

### LINKED ISSUE #579: provide a backoff for 403 errors
When Octant encounters 403 errors is should backoff

wwitzel3:
fixes #579 

### DIFF
--- a/changelogs/unreleased/579-wwitzel3
+++ b/changelogs/unreleased/579-wwitzel3
@@ -1,0 +1,1 @@
+Implements a backoff strategy for objectstore access.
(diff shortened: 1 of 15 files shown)


wwitzel3:
This solution doesn't quite work. If you attempt to visit an object that you have in the list cache after authorization goes away, it will attempt to start a watcher and you'll see the reflector.go errors begin to spam the console.

That said, maybe that is good enough? If you don't attempt to navigate anywhere after losing authorization, there is no API spam that happens.

The thought is, if a user is actively using Octant they would be able to correct / reload the context that is unauthorized.

The solution in #589 which implements the poller backoff does not suffer from this same problem.

wwitzel3:
This in combination with #592 should result in a much better experience for folks who are using Octant in an environment with rotating credentials.

bryanl:
Could you update the git changelog so we won't forget what this does?

bryanl (approved):

Issue #567, opened 2020-01-24
upgrade to Angular 9


wwitzel3:
Angular 9 at the time of creating this issue was in RC 10 and very very close to release. We want to move to Angular 9 when it is released.

rossfenrick:
@wwitzel3 Looks like Clarity has a new release that updates to Angular 9

[3.0.0-next.8 Changelog](https://clarity.design/news/3.0.0-next.8)

Issue #525, opened 2020-01-09
Replace YAML viewer with editor component


bryanl:

wwitzel3:
replace the existing YAML viewer with the Monaco editor in read-only mode

wwitzel3:
Ran in to a bunch of dep issues yesterday and threw my hands up. Reset my branch from master to day and just started over and it seems to be going much smoother. Hoping to have a PR up for this soon.

wwitzel3:
I'm working through some examples I found at https://github.com/microsoft/monaco-editor-samples

wwitzel3:
Progress. I need to work on the styling of it all and ensure it switches styles on Dark/Light switching.

![image](https://user-images.githubusercontent.com/27317/74971006-5cf2ec00-53ed-11ea-800a-4b43578b2832.png)

PR #609, opened 2020-01-31
upgrade to angular9-rc12 to unblock monaco integration

wwitzel3:
This PR allows us to start with #525 
We will continue to track the upcoming release of Angular9 and perform an upgrade to the official release when it is out. See #567 


wwitzel3:
my browser cache was preventing me from seeing that this is horribly broken, investigating

Issue #610, opened 2020-02-03
Rearrage angular modules to support web and app UIs


bryanl:
The move to the app will not take place all at once, so
to easy the transition, create a way to develop on the
web and app UIs of octant at the same time.

PR #612, opened 2020-02-03
Allow for web and app UIs to exist at the same time.

bryanl:
This change moves the existing application to a module
named `sugarloaf` and creates a stub module named
`denali` for the new app work.

#610


bryanl:
I found an issue. If I go to http://localhost:7777, I get empty content. If I then click on a navigation item, it loads up. I need to figure out why that is happening, but hopefully, that is the last issue.

wwitzel3 (approved):

wwitzel3 (commented):
Once that root / visiting issue is resolved, this looks good to me.

PR #628, opened 2020-02-04
fix tests on Windows

wwitzel3:

### DIFF
--- a/internal/kubeconfig/kubeconfig_test.go
+++ b/internal/kubeconfig/kubeconfig_test.go
@@ -36,7 +36,7 @@ func TestFSLoader_Load(t *testing.T) {
 
 	l := NewFSLoader()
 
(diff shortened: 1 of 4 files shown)


internal/kubeconfig/kubeconfig_test.go
```diff
@@ -36,7 +36,7 @@ func TestFSLoader_Load(t *testing.T) {
 
 	l := NewFSLoader()
 
-	kc, err := l.Load(strings.Join(paths, ":"))
+	kc, err := l.Load(strings.Join(paths, string(os.PathListSeparator)))
```
bryanl: Isn’t it already a string?

wwitzel3: No it's a Rune

PR #643, opened 2020-02-07
Remove boolean value from object store Get

### LINKED ISSUE #540: grpc: server failed to encode response
My plugin is generating the following error message in the console the first time that the user navigates to the plugin in the UI when the `Get` function in the dashboard client is executed:

`grpc: server failed to encode response:  rpc error: code = Internal desc = grpc: error while marshaling: proto: Marshal called with nil`

and after the first time navigating there, every invocation of the function works flawlessly. To reproduce the error:

1. modify the sample plugin in the octant source code, so that its handleNavigation and initRoutes functions (in main.go) are replaced with:

```
func handleNavigation(request *service.NavigationRequest) (navigation.Navigation, error) {
        return navigation.Navigation{
                Title:    "Sample Plugin",
                Path:     request.GeneratePath(),
                IconName: "cloud",
        }, nil
}

func initRoutes(router *service.Router) {
        router.HandleFunc("*", func(request *service.Request) (component.ContentResponse, error) {
                contentResponse := component.NewContentResponse(component.TitleFromString("Example"))

                ctx := request.Context()
                client := request.DashboardClient()
                key := store.Key{APIVersion: "apps/v1", Kind: "Deployment", Namespace: "argo", Name: "argo-ui"}

                _, _, err := client.Get(ctx, key)
                if err != nil {
                        contentResponse.Add(component.NewText("Failed"))
                } else {
                        contentResponse.Add(component.NewText("Success"))
                }
                return *contentResponse, nil
        })
}
```
2. Install argo with the two commands listed at https://github.com/argoproj/argo . (This is needed since the `Get` is trying to fetch the argo deployment)
3. Compile octant, install the plugin, and Launch octant
4. When you navigate to the Sample plugin you will see the error in the console. "Success" is displayed in the UI, indicating that things are fine. Navigate around and notice that the console error does not appear again.

GuessWhoSamFoo:
This PR removes the found boolean from Get because both an invalid key or non-existent object will return a nil object. The value is redundant and potentially misleading because Get uses the dynamic informer as a fallback.

This also fixes the case where a Get returns nil and prevents encoding of a nil object.

- Fixes #540

**Special notes for your reviewer**: This is a breaking API change :warning:

### DIFF
--- a/internal/objectstore/dynamic_cache.go
+++ b/internal/objectstore/dynamic_cache.go
@@ -315,22 +315,22 @@ type getter interface {
 }
 
 // Get retrieves a single object.
-func (dc *DynamicCache) Get(ctx context.Context, key store.Key) (*unstructured.Unstructured, bool, error) {
+func (dc *DynamicCache) Get(ctx context.Context, key store.Key) (*unstructured.Unstructured, error) {
 	ctx, span := trace.StartSpan(ctx, "dynamicCacheGet")
 	defer span.End()
 
 	if dc.isBackingOff(ctx, key) {
-		return &unstructured.Unstructured{}, false, nil
+		return &unstructured.Unstructured{}, nil
 	}
 
 	if err := dc.access.HasAccess(ctx, key, "get"); err != nil {
 		if meta.IsNoMatchError(err) {
-			return &unstructured.Unstructured{}, false, nil
+			return &unstructured.Unstructured{}, nil
 		}
 		if !dc.isBackingOff(ctx, key) {
 			dc.backoff(ctx, key)
 		}
-		return nil, false, fmt.Errorf("check access for get to %s: %w", key, err)
+		return nil, fmt.Errorf("check access for get to %s: %w", key, err)
 	}
 
 	span.Annotate([]trace.Attribute{
@@ -341,16 +341,15 @@ func (dc *DynamicCache) Get(ctx context.Context, key store.Key) (*unstructured.U
 	}, "get key")
 
 	object, err := dc.getFromInformer(ctx, key)
-
 	if err != nil {
 		if kerrors.IsNotFound(err) {
-			return nil, false, nil
+			return nil, nil
 		}
 
-		return nil, false, err
+		return nil, err
 	}
 
-	return object, true, nil
+	return object, nil
 
 }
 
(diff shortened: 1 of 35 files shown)


internal/objectstore/dynamic_cache.go
```diff
@@ -341,16 +341,19 @@ func (dc *DynamicCache) Get(ctx context.Context, key store.Key) (*unstructured.U
...
 		if kerrors.IsNotFound(err) {
-			return nil, false, nil
+			return nil, nil
 		}
 
-		return nil, false, err
+		return nil, err
+	}
+
+	if object == nil {
```
wwitzel3: I don't think this is needed?

pkg/plugin/api/client.go
```diff
@@ -100,25 +100,25 @@ func (c *Client) List(ctx context.Context, key store.Key) (*unstructured.Unstruc
 }
 
 // Get retrieves an object from the dashboard's objectStore.
-func (c *Client) Get(ctx context.Context, key store.Key) (*unstructured.Unstructured, bool, error) {
+func (c *Client) Get(ctx context.Context, key store.Key) (*unstructured.Unstructured, error) {
```
wwitzel3: I'm still thinking about this. We can make this entire change, fix the bug, but still keep the API for Plugin Get the same. Providing a `found` helper value.

Internally, I think it makes sense, since as you mentioned, we are using the dynamic client, but for the plugins, does it make sense to keep the `found` value? What do you think?

GuessWhoSamFoo: An edge case where the object is nil while found is true is if there is a layer for validating keys. In other words, a found helper can be used to communicate to plugin authors that a given key will always return a nil object. Opened #655 to implement if we reach a consensus on this.

wwitzel3 (approved):

internal/objectstore/dynamic_cache.go
wwitzel3: @GuessWhoSamFoo I think 352 - 354 can be deleted?

wwitzel3: remove 352-354

wwitzel3 (changes requested):
1 minor thing and then LGTM

wwitzel3 (approved):

PR #645, opened 2020-02-07
add Request interface to make plugin testing easier

### LINKED ISSUE #491: External plugins not unit testable
We are developing an octant plugin that does not reside in the octant source tree and it appears that unit testing portions of it are not feasible. For example, the handleFunc that the plugin supplies (which is returned by router.Match) is invoked with a service.Request object, and service.Request contains private fields (dashboardClient, baseRequest) that cannot be assigned to or even mocked out.

The basic problem is that interfaces and types that are part of the plugin API need to be completely public so that they can be developed and tested outside of the octant source tree.

wwitzel3:

Adds a Request interface for plugins to allow easy mocking/faking of the service.Request type. This allows plugins to properly unit test their route handlers when they are accessing things like the Context.

A sample plugin was created to test this new interface: https://github.com/wwitzel3/octant-actions-plugin

- Fixes #491 

**Special notes for your reviewer**: This is a breaking API change :warning:

cc: @GarySmith

### DIFF
--- a/build.go
+++ b/build.go
@@ -236,7 +236,11 @@ func installTestPlugin() {
 	dir := pluginDir()
 	log.Printf("Plugin path: %s", dir)
 	os.MkdirAll(dir, 0755)
-	pluginFile := fmt.Sprintf("%s/octant-sample-plugin", dir)
+	filename := "octant-sample-plugin"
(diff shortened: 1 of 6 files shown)


GuessWhoSamFoo (approved):
lgtm

PR #663, opened 2020-02-11
fix dynamic backoff test

wwitzel3:

The dynamic backoff test is too close to the real amount of time the backoff will wait. This means it sometimes fails on slower systems. Extend it by 300 to ensure that it passes on GitHub Actions CI runs.


bryanl (commented):
This bit of knob turning is fixing the symptom rather than fixing the test. Should this test be re-worked?

wwitzel3:
Agreed. I guess we don't actually need to test the After call, we can make an assumption that respects what we pass it.

I'll extract the timing function and actually check the time.Duration for incrementing values.

wwitzel3:
I admit the PR was made in frustration. I'll fix it up.

wwitzel3:
@bryanl updated, PTAL

bryanl (approved):

PR #718, opened 2020-02-20
Summary sections can be updated

### LINKED ISSUE #717: Replace summary sections
It should be possible to replace summary sections. This can
be used by plugins to replace existing content in config and
status sections in summaries for objects.

bryanl:
Update Summary component to allow its summary sections to
be updated. To update a section, add a new item with an
existing name.

This change will allow plugins to update existing config and
status items in the object summary. This is useful for
converting text components to link components.

Fixes: #717 

### DIFF
--- a/changelogs/unreleased/717-bryanl
+++ b/changelogs/unreleased/717-bryanl
@@ -1,0 +1,1 @@
+Summary sections can be updated
--- a/pkg/view/component/summary.go
+++ b/pkg/view/component/summary.go
@@ -5,7 +5,11 @@ SPDX-License-Identifier: Apache-2.0
 
 package component
 
-import "encoding/json"
+import (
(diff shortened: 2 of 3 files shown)


bryanl:
@vmware-tanzu/octant-reviewers ping!

wwitzel3:
I worried about what happens when every plugin wants to write their own custom message to a well defined section, it is basically what ever plugin gets there last wins.

bryanl:
> I worried about what happens when every plugin wants to write their own custom message to a well defined section, it is basically what ever plugin gets there last wins.

@wwitzel3 Yes. The last one written wins. There's not an easy way to stop it from happening other than public shaming.

wwitzel3 (approved):

Issue #646, opened 2020-02-07
Unmarshal components for plugins


GuessWhoSamFoo:
These are components that are missing from `pkg/view/unmarshal.go`. 

 - [x] button/button group #647 
 - [x] code
 - [x] donut chart
 - [x] ~extension - can a plugin add a tab to the extension?~
 - [x] error
 - [x] ~form~ - n/a in this case
 - [x] ~port? - requires a port forward state~
 - [x] single stat
 - [x] ~terminal~
 - [x] ~vertical bullet~ removed
 - [x] ~yaml~ - considered deprecated as it will be an editor soon

Some of these components such as terminal or port requires knowledge of octant's internal state which might not be good developer experience in the context of plugins.

wwitzel3:
Port and Terminal we won't be doing as part of this pass.

wwitzel3:
This is really a tracking issue, please add new issues and link to them in the list as we address each component.

GuessWhoSamFoo:
Is vertical bullet chart supported anymore? I can't find the frontend for it anymore. iirc, the donut chart is used instead for workloads

PR #723, opened 2020-02-21
Add Monaco editor component for YAML viewer.

### LINKED ISSUE #525: Replace YAML viewer with editor component

wwitzel3:
Replaces the YAML viewer with a read only Monaco editor view.

- Fixes #525


wwitzel3:
Current issues:
- [ ] preventDefault spam in console
- [ ] needs dark theme

wwitzel3:
@mklanjsek can you take a look at this branch when you get a chance? I'm stuck on these preventDefault errors that spam the console.

mklanjsek:
Sure thing @wwitzel3 I'll take a look today

mklanjsek:
This looks like most promising way to work around the console spam:
https://github.com/zzarcon/default-passive-events

The core of the issue is that on mouse move event Monaco editor is calling preventDefault() and Chrome is throwing error because the event handler was declared as a passive event handler. There is a long story about advantages of using passive events [here](https://github.com/WICG/EventListenerOptions/blob/gh-pages/explainer.md), but Chrome is certainly overreacting by spamming the console (It also appears that initial warning has been replaced by error at some point). Unfortunately there doesn't seem to be a way to prevent the console error messages except by globally removing the passive flag from all event handlers using approach outlined above.

It's interesting that this is not happening in simple editor usages, for example [here](https://alauda.github.io/ng-monaco-editor/demo/), so before jumping into global event overrides I am going to try and see if there is something particular in our setup that's indirectly making this happen. 

Note that this is Chrome issue, there are no errors reported inside Firefox.

Also worth noting is that we see the same issue in our Resource View where `cytoscape` component is behaving the same way

mklanjsek:
@wwitzel3 what was the motivation of using `ng-monaco-editor` vs this:
https://github.com/materiahq/ngx-monaco-editor

Not sure if that would fix the error spamming issue but it looks like `ngx-monaco-editor` has better support for module loading and Electron

wwitzel3:
@mklanjsek I just liked the abstractions in ng-monaco-editor better, they seemed more Angular-ish. But if the ngx package works better for us, I've no concerns about switching to use it.

mklanjsek:
Implemented the fix for passive events errors in browser console.  Not the most ideal solution but probably the only possible for this very tricky issue.  Note that this fixes the same problems in Resource Editor.

mklanjsek:
> @mklanjsek I just liked the abstractions in ng-monaco-editor better, they seemed more Angular-ish. But if the ngx package works better for us, I've no concerns about switching to use it.

We should probably keep `ng-monaco-editor` then and switch only if we run into problems with Electron.  I tested both and didn't see much difference at least with editor performance.

wwitzel3:
I don't really understand this error I'm getting in the preflight checks, digging in to it now.

wwitzel3:
I got stuck on this yesterday trying to figure out what exactly TypeScript wanted from me with respect to how to reference the options. So I switched over to looking at the Azure port forwarding. I'm back looking at this today.

@mklanjsek do you have any thoughts on why the pre-flight Karma check is complaining about my options?

mklanjsek:
> @mklanjsek do you have any thoughts on why the pre-flight Karma check is complaining about my options?

Not sure, but could it be something related to this?
https://stackoverflow.com/questions/48448832/karma-angular-4-no-provider-for-service

wwitzel3:
Ahh good find, your search fu beats mine. Thank you. (I'll blame DDG here, lol)

wwitzel3:
I'm still stuck trying to get these karma tests to pass. @mklanjsek @GuessWhoSamFoo @bryanl 
If you have some time to take a look at the failure and give any ideas, it would be very much appreciated.

mklanjsek:
> If you have some time to take a look at the failure and give any ideas, it would be very much appreciated.

Interesting, looks like chrome headless is having problems loading Monaco, I'll take a look!

web/src/app/modules/shared/components/smart/editor/editor.component.ts
```diff
@@ -0,0 +1,45 @@
...
+    return this.v;
+  }
+
+  value: string;
+  options: Options;
+
+  constructor() {
+    this.options = {} as Options;
+  }
+
```
GuessWhoSamFoo: Navigation from a tab to YAML causes a loss in dark theme. `ngOnInit` is needed to load the correct theme

pkg/view/component/base.go
```diff
@@ -13,6 +13,7 @@ const (
 	typeCodeBlock           = "codeBlock"
 	typeContainers          = "containers"
 	typeDonutChart          = "donutChart"
+	typeEditor              = "editor"
```
GuessWhoSamFoo: This can be done in a separate PR, but also needs an entry in `unmarshal.go` xr https://github.com/vmware-tanzu/octant/issues/646

web/src/app/modules/shared/components/fixPassiveEvents.ts
```diff
@@ -0,0 +1,48 @@
+/*
+ *  Copyright (c) 2020 the Octant contributors. All Rights Reserved.
+ *  * SPDX-License-Identifier: Apache-2.0
```
GuessWhoSamFoo: nit: mis-indented header

web/src/app/modules/shared/components/presentation/content-switcher/content-switcher.component.html
```diff
@@ -39,6 +39,9 @@
     <ng-container *ngSwitchCase="'loading'">
       <app-view-loading [view]="view"></app-view-loading>
     </ng-container>
+    <ng-container *ngSwitchCase="'editor'">
```
GuessWhoSamFoo: nit: let's get these in alphabetical order

bryanl (approved):
Will follow up on nits in subsequent PRs.

PR #733, opened 2020-02-26
proposal for combined logging

wwitzel3:

Proposal for updates to our logging.

Please leave feedback and suggestions.

### DIFF
--- a/proposals/20200224-combined-logging.md
+++ b/proposals/20200224-combined-logging.md
@@ -1,0 +1,72 @@
+# Combined Logging
+
+Support the ability to view all of the container logs for a given workload. Currently you must request each container
+log separately.
+
+## Goals
+ - Combine the logging output from multiple containers.
+ - Tag/mark which log lines are from which container.
+ - Provide a list of which container logs are included in the stream.
+ - Only send content when there are new log entries.
+
+## Non-Goals
+ - Server-side filtering
+
+## Log Streaming
+
+### Current implementation
+Our existing log "streaming" is done via a log endpoint with a JSON request. We explicitly register a container logs
+handler in `internal/api/api.go`.
+
+    s.HandleFunc("/logs/namespace/{namespace}/pod/{pod}/container/{container}", containerLogsHandler(ctx, a.dashConfig))
+
+When visiting the Logs tab we generate an XHR matching the URL pattern we've registered for container logs. This XHR
+request is generated using the selected container from the listed of containers. This list of containers is provided
+by Octant to the Log component. A user can only view a single containers logs at a time.
+
+Example request:
+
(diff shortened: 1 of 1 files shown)


mklanjsek:
This looks great!  Additionally, we should try to add as much metadata about each log line as possible. Having that level of detail provides much richer logging experience and filtering capabilities.  This is specially true for labels as some of the best logging platforms are based on label tagging/filtering, for example [Loki](https://grafana.com/oss/loki/)

proposals/20200224-combined-logging.md
```diff
@@ -0,0 +1,72 @@
...
+Support the ability to view all of the container logs for a given workload. Currently you must request each container
+log separately.
+
+## Goals
+ - Combine the logging output from multiple containers.
+ - Tag/mark which log lines are from which container.
+ - Provide a list of which container logs are included in the stream.
+ - Only send content when there are new log entries.
+
+## Non-Goals
```
bryanl: Another suggestion I want to offer is that is combined logging for a pod. Multiple pods are not in scope.

proposals/20200224-combined-logging.md
```diff
@@ -0,0 +1,72 @@
...
+Add a sentinel value of `AllContainers` to Octant and make this the default choice when requesting logs.
+
+Add a set of new actions to Octant to handle starting a log stream.
+
+Add a new EventType to Octant for log streams that the WebsocketService can subscribe to.
+
+Extend our existing log entry to include the container the log entry is from.
+
+#### Stream Aggregation
+Create a go routine that wraps the `Stream` call for each containers Octant will stream logs for.
```
bryanl: Is this a good opportunity to create a better logging interface? I’m not 100% sure what it should look like but the ideas seems to be in the right direction.

proposals/20200224-combined-logging.md
wwitzel3: I'm thinking about this some more. I'm going to consider what complexity is added if we say this is to get logs combined from any workload vs. for all containers of a pod. I'm also going to consider how we might do this for only pods, but in way that makes it extensible for the next improvement, which would Application logs.

proposals/20200224-combined-logging.md
wwitzel3: Yeah, I think we will ultimately end up with a different interface. I will walk through some ideas and update the proposal to reflect the new design.

I'm thinking this change will create an `internal/logging` module which will house the action handlers, streaming functions, and helpers.

bryanl (commented):
My thoughts after thinking about this are the goals specified need a strategy to go along with them under the proposal setting. How are you going to work towards each goal?

bryanl (approved):

bryanl (approved):

PR #754, opened 2020-03-05
Move base64 decoding inside container

GuessWhoSamFoo:
 - Bumps goreleaser version
 - Difference in base64 version may be causing the pipe to fail. This change will match the one used in drone

- Fixes https://github.com/vmware-tanzu/octant/actions/runs/49781696

### DIFF
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -3,8 +3,8 @@ on:
   schedule:
     - cron:  '0 0 * * *'
 jobs:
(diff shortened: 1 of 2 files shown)


bryanl:
I'm not sure this worked.

GuessWhoSamFoo:
I don't have access to the secret to double check the contents. Going to skip around base64 decoding and try https://github.com/goreleaser/goreleaser/issues/1234#issuecomment-553369444 instead.

Issue #755, opened 2020-03-06
False positive error in ingress resource viewer with wildcard domains


rubenv:
**What steps did you take and what happened:**
Create an ingress with a wildcard certificate:

```
Name:             example
Namespace:        development
Address:          
Default backend:  default-http-backend:80 (<none>)
TLS:
  example-cert terminates *.development.example.com
Rules:
  Host                             Path  Backends
  ----                             ----  --------
  backend.development.example.com  
                                      backend:80 (<none>)
  docs.development.example.com     
                                      apidocumentation:80 (<none>)
  web.development.example.com      
                                      web:80 (<none>)
  api.development.example.com      
                                      api:80 (<none>)
  objects.development.example.com  
                                      minio:9000 (<none>)
Annotations:
  cert-manager.io/cluster-issuer:               letsencrypt-production-dns
  kubernetes.io/ingress.class:                  nginx
  nginx.ingress.kubernetes.io/proxy-body-size:  512m
Events:                                         <none>
```

The resource viewer will falsely claim that there's a problem with this ingress:

```
example

    No matching TLS host for rule "backend.development.example.com"
    No matching TLS host for rule "docs.development.example.com"
    No matching TLS host for rule "web.development.example.com"
    No matching TLS host for rule "api.development.example.com"
    No matching TLS host for rule "objects.development.example.com" 
```

**What did you expect to happen:**
Should match the wildcard rule, this ingress is OK.

**Environment:**

- Octant version (use `octant version`): 
```
Version:  0.10.2
Git commit:  438ce3d52088c1e3e15aaf9817ba71e3c85255f5
Built:  2020-02-14T23:12:26Z
```
- Kubernetes version (use `kubectl version`):
```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.0", GitCommit:"ddf47ac13c1a9483ea035a79cd7c10005ff21a6d", GitTreeState:"clean", BuildDate:"2018-12-03T21:04:45Z", GoVersion:"go1.11.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.7", GitCommit:"c83d931fb9bece427bc63a02349755e0f8696d3e", GitTreeState:"clean", BuildDate:"2020-01-31T20:09:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
[...]

PR #756, opened 2020-03-06
Failing test case for #755

rubenv:
Test case for #755

None yet!


rubenv:
Checks obviously fail, but this illustrates the problem and should be a good starting point for fixing the issue.

bryanl:
closes in favor of #757

PR #759, opened 2020-03-09
Use Labels overflow pattern

### LINKED ISSUE #636: Use label overflow pattern
When there are lots of labels or selectors, it makes datagrid columns or cards much larger than they need to be. We should adopt the label overflow/badge pattern used in Clarity.

As-is:
Screen Shot 2020-02-05 at 11 39 34 AM

To-be:
Screen Shot 2020-02-05 at 11 37 51 AM

Screen Shot 2020-02-05 at 11 39 07 AM

The badge is clickable and when clicked a tooltip opens displaying the rest of the labels/selectors (which are also clickable). This will reduce visual clutter and extraneous whitespace.

nfarruggiagl:

Adopt the label overflow/badge pattern used in Clarity so when there are lots of labels or selectors, it doesn't make datagrid columns or cards much larger than they need to be anymore.

Before
Before

After
After

- Fixes #636 

Had to make some changes to Clarity styles to allow this behaviour since datagrids don't allow overflow content.

**Release note**:
```
release-note
```

### DIFF
--- a/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
+++ b/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
@@ -1,0 +1,29 @@
+<div class="labels-wrapper" *ngIf="labels">
+  <ng-container *ngFor="let label of showLabels; trackBy: trackByIdentity">
+    <span
+      *ngFor="let item of label | keyvalue; trackBy: trackByIdentity"
+      class="label label-light-blue clickable"
+      (click)="filterLabel(item.key, item.value)">
+      {{ item.key }}:{{ item.value }}
+    </span>
+  </ng-container>
+
+  <clr-signpost>
+    <span
+      *ngIf="overflowLabels && overflowLabels.length > 0"
+      class="badge badge-light-blue"
+      clrSignpostTrigger>
+      {{ overflowLabels.length }}+
+    </span>
+    <clr-signpost-content *clrIfOpen>
+      <ng-container *ngFor="let label of overflowLabels; trackBy: trackByIdentity">
+        <span
+          *ngFor="let item of label | keyvalue; trackBy: trackByIdentity"
+          class="label label-light-blue clickable"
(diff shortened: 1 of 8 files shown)


mklanjsek:
@nfarruggiagl, I am now seeing additional Info icon on fields that do not overflow (2nd row below). 
image

After clicking on it, an empty tooltip shows up:
image

Also note the expanded vertical space for these rows.

nfarruggiagl:
Indeed! Good catch @mklanjsek .. that should be gone now

web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
```diff
@@ -0,0 +1,29 @@
+<div class="labels-wrapper" *ngIf="labels">
+  <ng-container *ngFor="let label of showLabels; trackBy: trackByIdentity">
+    <span
+      *ngFor="let item of label | keyvalue; trackBy: trackByIdentity"
+      class="label label-light-blue clickable"
```
mklanjsek: I would recommend extracting out the the duplicating code segment that shows group of labels - notice that lines 2-9 are same as 18-25.

nfarruggiagl: I guess we could extract those lines into a child component and reuse it, but I think that would require more work and even more lines of code as it is right now.

rossfenrick:
@nfarruggiagl Sorry if this wasn't clear in the issue, but this should apply to all clarity label components, which includes both labels & selectors. It seems like it is only being applied to labels and not selectors.

nfarruggiagl:
@rossfenrick Oh! I see, we can add that on a different branch! I thought it was just for labels but we can apply the same on selectors as well

GuessWhoSamFoo:
For selectors, we have to be careful because it is the right-most column in a few datagrids and need to shift the signpost's position. This also affects timestamps in a few places.

Issue #664, opened 2020-02-12
Move Terminal component to tab


wwitzel3:
Currently the terminal component is created by executing a command using a button on the container in the pod listing.

Move the terminal window to its own tab. defaulting the command to be executed to /bin/sh and having a drop down to select which container the terminal is open against.

This to consider when implementing this:
 - what is /bin/sh is not found?
    - currently users specify the command they want to run, is this going away?
 - how long does the terminal stay open after navigating away without explicitly exiting?
 - if a user explicitly exits a terminal, does clicking away and back to the tab restart the session?
 - does navigating away from the tab exit the terminal session?
     - if exiting does end the session automatically, should we save session scrollback for later retrieval?

rossfenrick:
I'll have to think about this some more but another option would be moving the command to the view heading next to the delete button:
![Screen Shot 2020-02-12 at 11 26 02 AM](https://user-images.githubusercontent.com/7506366/74369689-a6f32680-4d8a-11ea-9ee1-72c1c0760e38.png)

gandazgul:
This would be awesome!

mklanjsek:
My initial thought was that `Terminal` should not be in top right corner where we only have immediate actions, like `Delete`, etc.  But to @rossfenrick point, it is an action after all so that may be a right place for it.

As fo other questions, I think we should start with the reasonable defaults and add things like command to run to global settings.

@wwitzel3 thoughts?

wwitzel3:
- There will be a Terminal tab.
- The default command will be `/bin/sh` - w
- When you click on the tab, if no session exists, a new terminal session will be created.
- The session will remain active as long as you remain on the pod.
- When you navigate away from the pod, the session will be closed.
- If you exit an active session `# exit` the backlog will remain until you reconnect OR navigate away from the pod.

wwitzel3:
The changes above also mean:

- No longer a need for  a terminals listing describer

GuessWhoSamFoo:
Paired with @mklanjsek a bit on this today.

 - Since there is only one tab per pod, terminal manager should support a key value store of pod names and terminal ID per namespace.
 - We tried to handle creating terminal instance only on the backend. This will not work because backend does not track active tab.

PR #775, opened 2020-03-16
Move terminal to tab

mklanjsek:
Terminal is now available in separate tab inside the Pod view.

Note: I tried to make changes so each browser window creates a separate terminal session, however that was proven to be pretty hard. Passing the session instance from client to backend is not such a big deal (that is implemented now) but bigger problem is that there is only a single channel used to communicate with pod. That, combined with the fact that terminal is partially created when Pod view is opened and then completely created only when Terminal tab is clicked makes it really hard to keep the track of terminal instances and states. We can attempt to do this in future but it will be much bigger surgery in terminal management code.


web/src/app/modules/shared/components/smart/terminal/terminal.component.ts
```diff
@@ -13,18 +13,15 @@ import {
...
   TerminalOutputStreamer,
 } from 'src/app/modules/shared/terminals/terminals.service';
 import trackByIdentity from 'src/app/util/trackBy/trackByIdentity';
-import {
-  TerminalDetail,
-  TerminalView,
-} from 'src/app/modules/shared/models/content';
+import { TerminalView } from 'src/app/modules/shared/models/content';
 import { WebsocketService } from '../../../services/websocket/websocket.service';
-import { SliderService } from 'src/app/modules/shared/slider/slider.service';
```
GuessWhoSamFoo: There is some more cleanup for slider service. Mainly removing the remaining usage in `container.component.ts` and extra margin in the scss file then the service can be removed entirely.

web/src/app/modules/shared/components/smart/terminal/terminal.component.ts
```diff
@@ -33,57 +30,59 @@ import { Subscription } from 'rxjs';
...
       this.term = new Terminal({
         logLevel,
         disableStdin,
       });
-
       this.initStream();
       this.enableResize();
       this.term.onData(data => {
-        if (this.view.config.terminal.active === true) {
+        if (active) {
```
GuessWhoSamFoo: What should happen if the user exits the terminal from tab?

internal/modules/overview/overview.go
```diff
@@ -259,38 +259,6 @@ func (co *Overview) Content(ctx context.Context, contentPath string, opts module
 
 func (co *Overview) extensionDescriber(path, namespace string, options describer.Options) (*component.Extension, error) {
```
GuessWhoSamFoo: `extensionDescriber` can also be entirely removed along with the generatorFunc in `generator.go.

internal/modules/overview/terminalviewer/terminalviewer.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	key, err:=  store.KeyFromObject(tv.object)
+	if err != nil {
+		return nil, err
+	}
+
+	container:= ""
+	if len(pod.Spec.Containers) > 0 {
+		container= pod.Spec.Containers[0].Name
```
GuessWhoSamFoo: Are we always defaulting to the first container in the pod? What if that container was an init container?

wwitzel3: I think we can check if the container is an init container and skip it?

web/src/app/modules/shared/components/smart/terminal/terminal.component.ts
mklanjsek: Based on https://github.com/vmware-tanzu/octant/issues/664#issuecomment-593667180, this is enforced on exit:
`If you exit an active session # exit the buffer will remain until you reconnect OR navigate away from the pod.`

Do you think we need to add additional user feedback?

GuessWhoSamFoo: Yeah, generalizing this more, there should be a notification that alerts the user if the session is no longer active since there are other failure modes (lost connection, container crash).

web/src/app/modules/shared/components/smart/terminal/terminal.component.ts
mklanjsek: Great point, removed

web/src/app/modules/shared/components/smart/terminal/terminal.component.ts
mklanjsek: Added more feedback, so for example if you type `exit` now, terminal will show `process exited...` message

internal/modules/overview/terminalviewer/terminalviewer.go
mklanjsek: I implemented the recommendation from Wayne but still seeing the above error sometimes.  Not sure exactly why, so we may need to address this again in future

wwitzel3:
Please rebase/rewrite the commit history as well. Doesn't need to be a single commit, but remove these merges and off-fix commits

Issue #671, opened 2020-02-12
Octant has a delay when clicking a link


bryanl:
When clicking a link, Octant delays for ~1 second before updating the view.

goals:
* Figure out where the delay is coming from
* Remove the delay

additional notes:
* This may be related to tables re-rendering themselves after the initial load

![Screen Recording 2020-02-12 at 6 50 22 AM mov](https://user-images.githubusercontent.com/240/74332606-1c97cc00-4d64-11ea-9412-ac3f288e7348.gif)

wwitzel3:
I see this behavior on other pages as well, it seems to do with the set content path call and our streaming loop.

Areas I'd look in to when starting on this:
- ensure set content path interrupts and sends new steam content (pretty sure it does)
- look at component rendering
- only send changes to stream

GuessWhoSamFoo:
Dived into this a little bit - currently suspect the problem is occurring on the backed. Tested this by turning off change detection for multiple components and the delay after clicking a link remains.

wwitzel3:
I'm wondering if there is some hiccup between when the SetContentPath cancels the context and when the stream sends out the refreshed / new data.

GuessWhoSamFoo:
Going to close this for now. We can reopen if responsiveness starts to decrease as features are added. Ideally we'd have benchmarking for various aspects of the code base.

PR #776, opened 2020-03-16
Use OnPush strategy for navigation component

GuessWhoSamFoo:
This PR separates change detection of the navigation component from all other components. An OnPush strategy that detects updates to navigation will prevent the component from re-rendering each second.

- Part of #671 

`JSON.stringify` works as an object comparison operator because the order of keys are sorted from the backend. The namespace switcher is moved one level higher as namespace updates should be part of a separate change detection.


web/src/app/modules/sugarloaf/pipes/filtertext/filtertext.pipe.ts
```diff
@@ -0,0 +1,14 @@
+import { Pipe, PipeTransform } from '@angular/core';
```
mklanjsek: New file, needs a header

PR #778, opened 2020-03-18
Labels/selectors will no longer overflow their containers

### LINKED ISSUE #770: Labels/selectors overflow their containers

Screen Shot 2020-03-12 at 11 14 06 AM

The labels should fit in their container. Probably should more aggressively use vertical space as well.

Version:  (dev-version)
Git commit:  df6e6084
Built:  2020-03-12T15:13:10Z

- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Mac Chrome

nfarruggiagl:

Labels/selectors will no longer overflow their containers

- Fixes #770 

Datagrids cells will decide the best disposition for the labels inside of it depending of available size

**Release note**:
```
release-note
```

### DIFF
--- a/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
+++ b/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
@@ -1,13 +1,16 @@
 <div class="labels-wrapper" *ngIf="labels">
-  <ng-container *ngFor="let label of showLabels; trackBy: trackByIdentity">
-    <span
-      *ngFor="let item of label | keyvalue; trackBy: trackByIdentity"
-      class="label label-light-blue clickable"
-      (click)="filterLabel(item.key, item.value)"
-    >
-      {{ item.key }}:{{ item.value }}
-    </span>
-  </ng-container>
+  <div class="labels-container">
+    <ng-container *ngFor="let label of showLabels; trackBy: trackByIdentity">
+        <span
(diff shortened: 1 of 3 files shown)


nfarruggiagl:
> This fixed the `Selector` column nicely, but the `Labels` column is now messed up (badge is overflowing), see here:
> image

Actually is not messed up, this is how Datagrid determined the best possible view for labels. It's using a column-direction because if it uses row-direction then will have the overflow issue again, I agree with you @mklanjsek , the label's position is odd, but is either that or overflowing the container :( .. maybe @rossfenrick has another idea ??

mklanjsek:
That does not look right to me - the badge looks weird in the separate row.

wwitzel3:
We want to add ellipsis and not overflow the container, we want to the row heights consistent.

nfarruggiagl:
With the latest changes, Labels & Badges will no longer wrap, they will always be in the same row to make row's height consistent as expected. If the names are too long, they will have ellipsis:

Screen Shot 2020-04-13 at 10 42 30

Note: Sometimes (if we open a signpost in the last row for example), the Datagrid will create a vertical scroll, this is due to the way Clarity's Datagrid work. I tried to find a workaround to avoid this but it was impossible to do it without breaking other parts of the Datagrid. I think until it supports overflowing without breaking, this is the best we can do.

Screen Shot 2020-04-13 at 10 42 55

mklanjsek:
Looking good with latest changes!

PR #784, opened 2020-03-21
Moved tooltips for timestamps to left if not enough space

mklanjsek:
Make sure the timestamp tooltips fit inside the view.  They will now show up on the left side if necessary.

Note: Tooltips were quite huge so I also reduced the tooltips size from large to medium.


wwitzel3:
I'm confused by the commit history on this, can you rebase it.

PR #806, opened 2020-03-30
Client side dependency bump

### LINKED ISSUE #788: Update client side dependencies
This is a good time to update all client side dependencies to latest version and verify everything works well.

mklanjsek:
What this PR does / why we need it:

- Updated all dependencies to latest versions,
- Fixed a few lint problems due to new linter rules,
- Addressed some unit test flakiness

Which issue(s) this PR fixes
Fixes #788


GuessWhoSamFoo:
Calling out more explicitly a part of this PR more explicitly, prettier 2.0 always defaults to using arrow function parentheses but it looks like we are keeping it to the 1.x behavior for the time being.

mklanjsek:
> Calling out more explicitly a part of this PR more explicitly, prettier 2.0 always defaults to using arrow function parentheses but it looks like we are keeping it to the 1.x behavior for the time being.

Right.  I decided to keep it as is for now as it would require quite a few changes, plus actually I like it more the way it is.

PR #810, opened 2020-04-01
Add e2e test proposal

GuessWhoSamFoo:

### DIFF
--- a/proposals/20200330-e2e-testing.md
+++ b/proposals/20200330-e2e-testing.md
@@ -1,0 +1,97 @@
+# Octant e2e Testing
+
+Create a stable end-to-end testing developer workflow. The previous iteration of e2e tests required a given cluster state with expected Octant UI performance before executing tests resulting in flakey tests.
+
+## Goals
+
+ - Mock backend data to avoid maintaining cluster state for tests
+ - Remove dependencies such as kubectl and kind needed to start e2e tests
+ - Provide an alternative for difficult testing cases
+ - Gain feedback for automated user interaction
+
+## Non-Goals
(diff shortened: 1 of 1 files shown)


mklanjsek:
Nicely done, I like the suggested approach!  
I would advocate creating the common e2e library or helper classes early on, to build the strong foundation for future tests. That will make creating new test as frictionless as possible which is one of the most important requirements for successful testing platform. 

How would running e2e tests locally interfere with local `kind` installation?

GuessWhoSamFoo:
If we switch to a different context or create namespaces, there needs to be a wrapper that ensures clean-up for other failures like syntax errors or if the previous process crashed and did not have a chance to clean up resources. The same bit of code starting creating the cluster resources for testing needs to check if it already exists in the event of an unhappy path.

wwitzel3:
I've read this over a couple times and I like it. I want to take another pass at it after a weekend away from it.

Some more insight in to what I'm thinking about as I read this:

One of the mistakes I made in my logging proposal is that my implementation strategy was a bit too naive, meaning, I wasn't really able to take the strategy and directly create a cascade of dependent tickets easily.

 I'm going to read this again but with the mindset of how would I create issues based of this proposal and what do they look like, how big are they? The goal of these proposals is to give us time to ingest the information so that we create an actual strategy (as ready to work issues).

wwitzel3:
This looks good. We can create an Epic to track this work and create / add any issues to it. We will want to create issues that map as directly as possible to the implementation strategy in this proposal.

Nice job @GuessWhoSamFoo

PR #812, opened 2020-04-03
Vertical Navigation 

mklanjsek:
After ironing a few layout issues with @rossfenrick, I think this is a got point to be reviewed and merged to master. Vertical navigation now works for both expanded and collapsed state with the hand-rolled flyout menu. Unfortunately, there was some history lost during merge from the old branch due to so many changes since that work started.  

@vmware-tanzu/octant-reviewers feedback please!

### DIFF
--- a/web/src/app/modules/sugarloaf/components/smart/container/container.component.scss
+++ b/web/src/app/modules/sugarloaf/components/smart/container/container.component.scss
@@ -2,50 +2,52 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-$navigation-width: 13rem;
-
 .main-container {
   // Navigation's color variables for Light Theme.
   :host-context(body) {
     --navigation-border-color: rgb(215, 215, 215);
-    --navigation-background-color: #0079ad;
+    --navigation-background-color: #0079AD;
   }
 
   // Navigation's color variables for Dark Theme.
   :host-context(body.dark) {
-    --navigation-border-color: #0e161b;
-    --navigation-background-color: #0e161b;
+    --navigation-border-color: #304250;
+    --navigation-background-color: #304250;
+  }
+
+  .header-centered {
+    margin: auto 0;
   }
 
   .header {
     background: var(--navigation-background-color);
 
-    .branding {
-      min-width: $navigation-width;
-    }
-
     .header-nav {
       padding-left: 1rem;
 
+      .namespace-switcher,
       .input-filter {
         margin-right: 10px;
+      }
+
+      .namespace-switcher {
+        display: flex;
+        flex-direction: column;
+        justify-content: center;
+        align-items: center;
+        margin-right: 20px;
+      }
+
+      .input-filter {
         z-index: 999; // clarity's data-grid beats this
         width: 300px;
       }
-    }
-
-    .header-actions {
-      display: flex;
-      align-items: center;
     }
   }
 
   .content-container {
     .navigation {
       border-right: 1px solid var(--navigation-border-color);
-      overflow-y: auto;
-      min-width: 2rem;
-      width: $navigation-width;
     }
 
     .content-area {
--- a/web/src/app/modules/sugarloaf/components/smart/content/content.component.html
+++ b/web/src/app/modules/sugarloaf/components/smart/content/content.component.html
@@ -1,4 +1,4 @@
-<div class="content-component">
+<div clrFocusOnViewInit class="content-component">
   <ng-container *ngIf="hasReceivedContent">
     <ng-container
       *ngIf="hasTabs; then withTabs; else withoutTabs"
--- a/web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
+++ b/web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
@@ -1,10 +1,27 @@
 <div class="app-namespace-selector">
-  <span class="title">Namespace</span>
-  <ng-select
-    class="namespace-dropdown"
-    [items]="namespaces"
-    (change)="selectNamespace($event)"
-    [(ngModel)]="currentNamespace"
-    [clearable]="false"
-  ></ng-select>
+  <ng-template [ngIf]="showDropdown()" [ngIfElse]="hideDropdown">
+    <clr-dropdown [clrCloseMenuOnItemClick]="true">
+      <button type="button" clrDropdownTrigger>
+        <clr-icon shape="namespace"></clr-icon>
+      {{ currentNamespace }}
+        <clr-icon shape="caret down"></clr-icon>
+      </button>
+      <clr-dropdown-menu *clrIfOpen [clrPosition]="'bottom-right'">
+        <label class="dropdown-header">Namespaces</label>
+        <ng-container *ngFor="let namespace of namespaces;trackBy: trackByIdentity">
+          <button
+            type="button"
+            [ngClass]="namespaceClass(namespace)"
+            clrDropdownItem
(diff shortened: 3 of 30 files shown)


web/src/app/modules/sugarloaf/components/smart/container/container.component.scss
```diff
@@ -2,50 +2,54 @@
...
     }
-
-    .header-actions {
-      display: flex;
-      align-items: center;
-    }
   }
 
   .content-container {
+    margin-bottom: 1.5rem;
```
GuessWhoSamFoo, line 49: Why the the content-container need a margin here?

web/src/app/modules/sugarloaf/components/smart/content/content.component.html
```diff
@@ -1,4 +1,4 @@
-<div class="content-component">
+<div clrFocusOnViewInit #scrollTarget class="content-component">
```
GuessWhoSamFoo, line 1: `scrollTarget` was removed as part of slider cleanup in https://github.com/vmware-tanzu/octant/commit/bcf1148c0f23811866969e124464a0ba3b093fd7#diff-4399b3274650f85b70c641d7e49db84d

web/src/app/modules/sugarloaf/components/smart/navigation/navigation.component.html
```diff
@@ -1,58 +1,106 @@
...
+        </ng-container>
+
+        <ng-template #noChildCollapsed>
+              <span class="tooltip-override tooltip tooltip-md">
+                <a clrVerticalNavLink
+                   [ngClass]= "{'item-selected' : i == this.lastSelection}"
+                   [routerLinkActiveOptions]="{exact: true}"
+                   [routerLink]="formatPath(section.path)"
+                   routerLinkActive="active"
+                   (mouseover)="flyoutIndex= -1; lastSelection= i"
```
GuessWhoSamFoo, line 45: Upon moving the mouse away from the collapsed nav, the item should not remain highlighted

web/src/app/modules/sugarloaf/components/smart/navigation/navigation.component.html
```diff
@@ -1,58 +1,106 @@
...
+            </div>
+          </a>
+        </ng-template>
+      </ng-template>
+    </ng-container>
+  </div>
+
+  <!-- Theme switcher -->
+  <div class="theme-switcher">
+    <app-theme-switch-button [collapsed]="collapsed"></app-theme-switch-button>
```
GuessWhoSamFoo, line 104: nit: The theme switch button in a collapsed navigation should remain on the bottom rather than shifting upwards as a result of collapsed content.

web/src/app/modules/sugarloaf/components/smart/navigation/navigation.component.scss
```diff
@@ -2,18 +2,147 @@
...
+
+  .flyout-group {
+    display: flex;
+    flex-direction: column;
+  }
+
+  .flyout {
+    display: flex;
+    border: 1px solid var(--navigation-border-color);
+    border-left-width: 0px;
```
GuessWhoSamFoo, line 112: It looks odd without a left border

web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
```diff
@@ -1,10 +1,27 @@
...
+            (click)="selectNamespace(namespace)">
+          {{ namespace }}
+          </button>
+        </ng-container>
+      </clr-dropdown-menu>
+    </clr-dropdown>
+  </ng-template>
+  <ng-template #hideDropdown>
+    <clr-icon shape="namespace"></clr-icon>
+    [All Namespaces]
```
GuessWhoSamFoo, line 25: Hit a case where navigation from `/cluster-overview` to `/` via top left octant logo results in stuck on `All Namespaces`

web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
```diff
@@ -1,10 +1,27 @@
...
+  </ng-template>
```
GuessWhoSamFoo, line 26: Another case where being in `/cluster-overview` and expanding a namespaced navigation section changes the namespace.

GuessWhoSamFoo:
As discussed in slack, we should merge this to master and break apart pending issues into an epic with the main one being losing a level of hierarchy for plugins.

web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
mklanjsek: I think this is OK - showing the namespace properties is not equivalent to switching to that namespace.

mklanjsek:
All good points, thanks @GuessWhoSamFoo!

web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
mklanjsek: At the end I decided to follow your recommendation and switch the namespace in this case.

web/src/app/modules/sugarloaf/components/smart/content/content.component.html
mklanjsek: Removed

web/src/app/modules/sugarloaf/components/smart/navigation/navigation.component.html
mklanjsek: This was bugging me too, finally figured out how to do it

web/src/app/modules/sugarloaf/components/smart/namespace/namespace.component.html
mklanjsek: Done

rossfenrick:
>  The sidebar should be collapsed by default. (we may even consider not allowing expansion at all later)

I'd highly recommend against this. A widely accepted UX best practice is that primary navigation should only be hidden if it the viewport is constrained by size (mobile).

bryanl:
@rossfenrick The idea is to make it act more like vscode's sidebar. It would not be hidden, but would instead take up less room.

rossfenrick:
@bryanl Sorry, when I say "hidden" I mean collapsed/only icons. Something we could change is if the user chooses to collapse the navigation, it stays like that on refresh and the next time they start the app. 

What's the benefit to removing the ability to expand & collapse?

mklanjsek:
Expanded/collapsed state should certainly be a user level setting stored in browser local storage, similar to theme and other preferences.

As for default state, expanded state is certainly geared towards new and novice users, so from that point of view it's probably a good choice to be a default state. If we can remember that state once a user flips the switch, we should be able to handle wide spectrum of user affinities.

wwitzel3:
I see lots of tools and other products that don't even offer an expanded state. They use tooltips / etc.. and our flyout menu + tooltip provide all of the information a person needs to navigate and they have a set of medium/large icons for the navigation.

I really like the idea of moving away from the expanded navigation all together, similar to tools like VS code. Making the collapsed view the only view and making the icons bigger.

mklanjsek:
I suggest we switch to collapsed, test that internally and make final decision before the next release.  Anybody against that?

mklanjsek:
I addressed all above issues, except this one:

> Namespace Overview and Cluster Overview both just read Overview and they are ambiguous when the menu is collapsed.

Unfortunately client does not have enough info to show that in tooltip. Change will be required on backend to add tooltip text, so I'll open a separate issue to handle that..

Issue #808, opened 2020-04-01
Plugins need a way to know the currently selcted namespace


bryanl:
Plugins can act like Octant modules, and they will need a way to know 
the current namespace. In my current work, I want my plugin to be
able to display a summary of certain objects within a namespace.

In Octant, the `ModuleProxy` implements the `Module` interface. `Module`
already includes a `SetNamespace` function that is called when the 
namespace changes. For `ModuleProxy`, this item is set as a no-op. To
make this work, the `ModuleService` interface needs a new entry 
`SetNamespace` that will in turn use the plugin machinery to call
this function on the plugin.

bryanl:
I wonder if it is possible to do without requiring an API change. The current plugin service API will now allow it.

GuessWhoSamFoo:
There are two cases here: getting the current namespace and displaying objects of a given namespace.

Tested that it is possible to display objects from a different namespace via `DashboardClient.Get` as long as the key is valid but it will require conversion from an unstructured object.

I always thought the way of getting the current namespace is from `service.PrintRequest.Object`. There should be a more obvious way via `CurrentNamespace`

wwitzel3:
Bryan and I talked about this today and decided that we could use actions to handle this.

I will have a PR up shortly that will namespace all of our events and actions.

After that, we can determine the best way to ensure that a set of actions are also sent to plugins. In this case the `action.octant.dev/setNamespace` action.

PR #820, opened 2020-04-06
Add namespacing to actions and events

wwitzel3:

This is the first step in #808


GuessWhoSamFoo (approved):
I haven't heard of anyone using octant events before but this is potentially api breaking for plugins.

PR #831, opened 2020-04-09
load summary even if actions fail to load

wwitzel3:

Summary tab fails to load because generating actions will can the object path function which produces an error for disconnected pods.

https://github.com/vmware-tanzu/octant/blob/master/internal/printer/container.go#L508

Log and continue.

### DIFF
--- a/changelogs/unreleased/831-wwitzel3
+++ b/changelogs/unreleased/831-wwitzel3
@@ -1,0 +1,1 @@
+Fix bug that prevented Pod summary loading if status was nil/unknown.
--- a/internal/printer/container.go
+++ b/internal/printer/container.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/vmware-tanzu/octant/internal/log"
 	"path"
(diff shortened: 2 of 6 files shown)


GuessWhoSamFoo:
This can end up spamming the console. Can it check for the parent pod status and decide to add the action func instead?

wwitzel3:
Yes.

wwitzel3:
@GuessWhoSamFoo so an issue I run in to when doing that.

Pod Templates which are part of a Deployment summary also use the container configuration printer.
I'm still trying to decide the best way to handle this.

wwitzel3:
I pushed up a new effort. Now the pod template is explicitly adding the action. No if/else song and dance. I left the error message in for those edge cases, but it no longer spams.

wwitzel3:
@vmware-tanzu/octant-reviewers PTAL

GuessWhoSamFoo (approved):
lgtm, this is great!

wwitzel3:
Thanks @GuessWhoSamFoo , added a changelog entry, will merge after checks

PR #834, opened 2020-04-10
Breadcrumbs 

mklanjsek:
Added breadcrumbs to page headers based on design from Ross defined [here](https://www.figma.com/file/CvMwJPlvETtFoPx43oXKUd/Breadcrumbs-and-back-button?node-id=1%3A152).


wwitzel3:
sorry for the delay, reviewing this now

wwitzel3:
https://youtu.be/z9kAMoNLBkA

mklanjsek:
> When viewing Workloads the breadcrumb navigation becomes confusing. The breadcrumb takes me back to a completely different view.

The breadcrumbs reflect the object structure, so for example a pod will always have `Pods` node as its parent. 
What you are expecting is that breadcrumbs work the same way as browser history, where you can go back and forth from one object in history to another regardless of their object hierarchy. That is a totally different paradigm and we should not mix these two.

@wwitzel3 is your expectation how this is supposed to work based on using Octant in browser for a long time?  If it wasn't for that, would object hierarchy based breadcrumb paradigm make sense?

wwitzel3:
My expectation is that my breadcrumbs would take me back to where I was.

So I understand what makes this weird. When I'm looking at Workloads / Deployments, I'm viewing two levels of hierarchy at the same time and then clicking into a deployment is three levels deep.

But if I'm only looking at Deployments, then I click a deployment, I'm two levels deep.

I think I would be ok with something like:

(clickable)Workloads -> Deployments
(clickable)Deployments -> my-deployment

So that, when I click a deployment from the workloads view, without the need to dive back in to the menu or use a browsers back button, I can get back to the Workloads view I was originally on when I clicked a deployment.

mklanjsek:
Darn.

I think problem here is that we are trying to merge breadcrumb and header functionality inside a single UI component. In order for that component to behave as a header, we are trying to remove some nodes in our 3 level deep structure. I am realizing now that there is no good way to do that - Wayne's approach is more accurate reflection of object structure and current approach focuses more on usability and eliminates information that's obvious for most users. Neither one is ideal, so we may as well byte the bullet and show all 3 levels even if we need to use ellipses on some of the nodes.

@rossfenrick any thoughts on your side?

rossfenrick:
If we only want to display 2 levels deep, then we should not use breadcrumbs, but instead a back button or a switcher.

If we want to display >2 levels deep, we could separate out the breadcrumbs and the header, like so: 
Screen Shot 2020-04-15 at 11 23 12 AM
Screen Shot 2020-04-15 at 11 20 21 AM

wwitzel3:
I like the separation of breadcrumbs and header

mklanjsek:
Thank you so much @rossfenrick, I'll try that

wwitzel3:
How's this coming along, I imagine this ended up being a bit of work since it fundamentally changes how we render the top of every resource list and resource summary view.

If you run in to any gotchas or have questions, let me know.

mklanjsek:
@wwitzel3 thanks, getting close.  You are right, it did require considerable changes, but it will be a huge improvement.

mklanjsek:
Implemented latest design based on [above comment](https://github.com/vmware-tanzu/octant/pull/834#issuecomment-614202342)

Please let me know what you think. It's not always obvious what the root breadcrumbs should be, hope I made the right choices. Also, please try to tests as many resource objects as you can, I probably don't have all of them in my configuration.

wwitzel3:
I think this is good.

Is it weird to have breadcrumbs AND a back button? lol .. reason I ask, is this example.

- View: Workloads -> Deployments -> deployment-foo
- Click: pod-bar
- View: Workloads -> Pods -> pod-bar
- Not what I wanted, back to the Deployment ? ...

wwitzel3:
The duplication of the breadcrumb and heading when on the root view is something we can think about dealing with later, for now it is fine.

mklanjsek:
I've been bugged by duplication of the breadcrumb and heading as well, but not sure if we can do much about it, at least I couldn't find acceptable solution so far. Certainly something to think about...

rossfenrick:
This is close, I think we just need to make a few changes. In general I think we want to be more object heirarchy based than navigation based. In Wayne's example, the pod `pod-bar` is a child to the deployment `deployment-foo`, and the breadcrumbs should reflect that.

Examples:
Select workloads and then deployments
`Workloads -> Deployments`

Then select an individual deployment
`Workloads -> Deployments -> deployment-1`

Then select an individual pod
`Deployments -> deployment-1 -> pod-1`

----

Select Jobs
`Workloads -> Jobs`

Then select job-1
`Jobs -> job-1`

Then select  pod-1
`Jobs -> job-1 -> pod-1`

----

We could also add a 4th level of breadcrumb links for an individual pod

Select pod-1
`Workloads -> Deployments -> deployment-1 -> pod-1`

mklanjsek:
@rossfenrick I am confused: how can `pod-1` in your example have two different parents (`job-1` and `deployment-1`)? That in my mind reflects navigation path and not object hierarchy.

It's important not to mix these two different breadcrumb types. I think we want to use attribute based breadcrumbs  as defined [here](https://www.smashingmagazine.com/2009/03/breadcrumbs-in-web-design-examples-and-best-practices/) or [here](https://medium.com/madison-ave-collective/are-breadcrumbs-still-fresh-for-ux-6e72b0f96e9b) That is the structure we have right now.

Having said that, I am still not sure our hierarchy is defined properly for all pages, so for example your 4 levels example may make more sense than the current implementation.  I think we need to go through all our pages and make sure breadcrumbs make sense and are consistent across the board.

rossfenrick:
Oh, they don't. Those are different examples (different pods belonging to different objects). Sorry for the confusion

PR #844, opened 2020-04-15
Allow plugins to handle Octant actions.

### LINKED ISSUE #808: Plugins need a way to know the currently selcted namespace

wwitzel3:
Adds the ability for events to have multiple action handlers registered.
This allows for our `Dispatch` method to call each registered handler which allows plugins to register handlers for our public Octant actions (currently just one, `action.octant.dev/setNamespace`).

Updates the service.ActionRequest object to now include an `ActionName` which is populated
during the `Dispatch` call. Allowing plugins to easily switch on the ActionName to handle them as desired.

- Fixes #808

**BREAKING PLUGIN API CHANGE**


wwitzel3:
Demo video of the sample plugin running on this branch.

https://youtu.be/GooC7W5L59M

GuessWhoSamFoo (approved):

PR #858, opened 2020-04-22
Remove slider and yaml components; added more onpush change detection

GuessWhoSamFoo:
Removes unused components and changes the content component to trigger change detection only when content is loaded


wwitzel3:
I have a branch that is using the extension view with a datagrid to hold errors.

GuessWhoSamFoo:
I'm going to close this for now. It's not clear there are any performance improvements. Removing the YAML component can be done in a different PR.

PR #861, opened 2020-04-22
Slow tabs

bryanl:

This change does two things:
1. In the frontend, redirect only when needed. Before, redirects were happening on every content update. This change sets it to only redirect when the content path has changed.
1. In the backend, consolidate tab printing for custom resources and non-custom resources. This change also runs the tab generate process concurrently.

**Release note**:
```
Consolidate tab printing functions
```


internal/describer/object.go
```diff
@@ -49,82 +51,80 @@ type Object struct {
...
+		iconName:           c.IconName,
+		iconSource:         c.IconSource,
+		tabsGenerator:      tg,
+		tabFuncDescriptors: td,
+	}
+
+	o.tabFuncDescriptors = []Tab{
+		{Name: "Summary", Factory: SummaryTab},
+		{Name: "Metadata", Factory: MetadataTab},
+		{Name: "Resource Viewer", Factory: ResourceViewerTab(false)}, // TODO: Fix me
```
wwitzel3, line 87: can we create an issue for this fix me with what needs to be fixed, can leave the TODO but have it reference a GH issue.

internal/describer/tab.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/vmware-tanzu/octant/internal/log"
+	"github.com/vmware-tanzu/octant/internal/modules/overview/logviewer"
+	"github.com/vmware-tanzu/octant/internal/modules/overview/terminalviewer"
+	"github.com/vmware-tanzu/octant/internal/modules/overview/yamlviewer"
+	"github.com/vmware-tanzu/octant/internal/printer"
+	"github.com/vmware-tanzu/octant/internal/resourceviewer"
+	"github.com/vmware-tanzu/octant/pkg/view/component"
+)
+
+func CustomResourceSummaryTab(crdName string) TabFactory {
```
wwitzel3, line 26: I'd prefer to see all of these share the same signature. I'd expect them all to return a TabFactory, not some of them also directly be a TabFactory based on the naming convention. My initial thought is to make them all return a TabFactory and create a TabFactoryOptions type they can accept for the factories that need setup before we do Component generation.

wwitzel3 (approved):
Nice refactor, couple comments that aren't blocker for the merge.

Issue #882, opened 2020-04-25
component.NewMarkdownText() renders badly in Table + SummarySection


jstrachan:
using the Markdown text component is a nice way to add formatted text, external links to documentation/github and whatnot and render clarity badges to improve the UX.

However when switching from Text -> MarkdownText the table/summary sections render with an extra line of whitespace. Am guessing this is CSS related rather than intentional?

e.g. here's a SummarySection with 2 fields; one Text, one MarkdownText. The same happens if using the MarkdownText in a Table cell:

<img width="324" alt="Octant" src="https://user-images.githubusercontent.com/30140/80277577-e8d40d80-86e7-11ea-8541-51b4f767f60a.png">

jstrachan:
I guess another option is to have a `HTMLText` component which allows some HTML markup inside (e.g. bold/italic/clarity badges and some external `<a>` tags with no JavaScript or extra magic?

jstrachan:
it looks like `<p>` inside of `<div class="markdown">` has a margin-top of `24` px which cause the layout issue. Haven't figured out a good scss PR yet, I'm a bit rusty on scss + angular

jstrachan:
aha - figured out the CSS fix; PR on its way shortly

PR #883, opened 2020-04-27
fix: layout MarkdownText better in tables/cards

jstrachan:

fixes layout so that a `component.NewMarkdownText` can be used in a table cell or card value

- Fixes #
#882


mklanjsek (commented):
Lgtm, thanks for doing that @jstrachan!

mklanjsek (approved):

PR #885, opened 2020-04-27
Enable editing of objects

bryanl:

This change adds a preview of an editor for object YAML

### DIFF
--- a/internal/api/container_logs.go
+++ b/internal/api/container_logs.go
@@ -8,11 +8,12 @@ package api
 import (
 	"context"
 	"fmt"
-	"github.com/vmware-tanzu/octant/internal/gvk"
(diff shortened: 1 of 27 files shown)


wwitzel3:
Taking a look this morning.

GuessWhoSamFoo:
image

Case: edit a pod's container environment variable (which is not possible). The error message can be cleaner with the "Forbidden: pod updates may not change fields other than...". It'd be even better if those lines can be blocked from editing, but this is not possible in monaco right now (https://github.com/Microsoft/monaco-editor/issues/953)

GuessWhoSamFoo:
We should catch ctrl + s key presses while the editor component is up. I am programmed to save reflexively after each edit

wwitzel3 (approved):

PR #907, opened 2020-05-02
Add delete action to object lists

bryanl:

This change adds delete actions to object lists (not CRDs).

Screen Recording 2020-05-02 at 1 50 13 PM mov

**Release note**:
```
Add a delete action to object lists
```

### DIFF
--- a/internal/describer/object.go
+++ b/internal/describer/object.go
@@ -162,7 +162,7 @@ func (d *Object) Describe(ctx context.Context, namespace string, options Options
 			return component.EmptyContentResponse, err
 		}
 
-		confirmation, err := DeleteObjectConfirmation(currentObject)
+		confirmation, err := octant.DeleteObjectConfirmationButton(currentObject)
 		if err != nil {
 			return component.EmptyContentResponse, err
 		}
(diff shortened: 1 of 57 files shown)


rossfenrick:
The button text in the modal could be more of a call-to-action, e.g. using "DELETE" instead of "OK".

Also when you delete an object from the object view (the page with summary, metadata, logs, etc.. tabs) the delete button is red. In that modal we use blue buttons and also in this modal we use a blue button and then a blue alert at the top of the page. We should try to be consistent and use either all red or all blue. Generally Clarity recommends using red for deletion/destructive actions.

bryanl:
These are good points. Will need to make some adjustments to enable Octant to generate these, but it’s all possible.

bryanl:
I took at a look Clarity at the clarity docs and made it possible to change the OK to Delete and the color depending on the action type.

bryanl:
Screen Shot 2020-05-06 at 9 56 03 AM

wwitzel3 (approved):

rossfenrick:
@bryanl didn't get to see this before it got merged in but one last thing: the DELETE button in the modal should be a solid primary button, not an outline button. i can make an issue for this if needed

bryanl:
@rossfenrick Yes make an issue. The other delete item needs to be updated as well.

PR #915, opened 2020-05-06
Fix Timestamp tooltips and overflow elements problems

mklanjsek:
Timestamp tooltips and overflow elements had some issues that are now addressed:

- Fixed Timestamp tooltip visibility issues,
- Converted Timestamp tooltip to use new tooltips like elsewhere,
- Stopped flipping the Timestamp tooltip - improved layout of new toolltip eliminates need for that,
- Fixed problem with selector/label overflow elements throwing exceptions on startup,   

### DIFF
--- a/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
+++ b/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.html
@@ -21,7 +21,7 @@
     </ng-container>
   </div>
 
-  <clr-signpost *ngIf="overflowLabels && overflowLabels.length > 0">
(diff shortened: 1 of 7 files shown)


GuessWhoSamFoo:
We need to have better visibility into karma's error logs, but this test needs a fix https://github.com/vmware-tanzu/octant/pull/915/checks?check_run_id=651047870#step:4:352

mklanjsek:
> We need to have better visibility into karma's error logs, but this test needs a fix https://github.com/vmware-tanzu/octant/pull/915/checks?check_run_id=651047870#step:4:352

Good catch, @GuessWhoSamFoo - need to make a habit of look closer into these. Anyway, should be fixed now

PR #918, opened 2020-05-07
fix: add a couple of CLI arguments

jstrachan:

adds a few helper CLI arguments:

* `browser-path` so you can specify the initial context path to open the browser as you start Octant.

* `accept-local-ip` so that we can pre-populate the accepted hosts with all the local IP addresses on your local network so you can use another local device to access octant. I've found this particularly useful so I can keep an eye on octant running on my laptop while I'm around the house on my iPhone ;)

**Release note**:
```

This release adds 2 new CLI arguments:

* `browser-path` so you can specify the initial context path to open the browser as you start Octant

e.g. to open octant on the Deployments view:

    octant --browser-path "/#/overview/namespace/default/workloads/deployments"

* `accept-local-ip` so that we can pre-populate the accepted hosts with all the local IP addresses on your local network so you can use another local device to access octant. 

e.g. if you start octant via:

    octant --listener-addr=0.0.0.0:7777 --accept-local-ip

You will be able to connect to http://your-laptop-ip-address:7777 from your phone/tablet on your local LAN.

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,16 @@
+
+
+all: web-build update
+
+build:
+	go run build.go build
+	cp build/octant /Users/jstrachan/bin/octant
+
+update:
+	go generate ./web
+	go run build.go build
+	cp build/octant /Users/jstrachan/bin/octant
+
+web-build:
+	cd web && npm run build
+	go generate ./web
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -26,8 +26,14 @@ import (
 
 const (
 	// ListenerAddrKey is the environment variable for the Octant listener address.
-	ListenerAddrKey  = "listener-addr"
+	ListenerAddrKey = "listener-addr"
+
+	// AcceptedHostsKey the list of accepted hosts to be allowed to connect over web sockets
 	AcceptedHostsKey = "accepted-hosts"
+
+	// AcceptLocalIPKey if no accepted hosts are specified this flag will add all the local IP addresses
+	//starting with `192.168.1.` to the accepted hosts
+	AcceptLocalIPKey = "accept-local-ip"
 	// PathPrefix is a string for the api path prefix.
 	PathPrefix          = "/api/v1"
 	defaultListenerAddr = "127.0.0.1:7777"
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,16 @@
+
```
wwitzel3, line 1: Looks like this got accidentally added :)

internal/api/api.go
```diff
@@ -26,8 +26,14 @@ import (
...
 const (
 	// ListenerAddrKey is the environment variable for the Octant listener address.
-	ListenerAddrKey  = "listener-addr"
+	ListenerAddrKey = "listener-addr"
+
+	// AcceptedHostsKey the list of accepted hosts to be allowed to connect over web sockets
 	AcceptedHostsKey = "accepted-hosts"
+
+	// AcceptLocalIPKey if no accepted hosts are specified this flag will add all the local IP addresses
+	//starting with `192.168.1.` to the accepted hosts
```
wwitzel3, line 35: This is too specific, anecdotally, I use `192.168.2.`

wwitzel3:
Initial thoughts

Here: https://github.com/vmware-tanzu/octant/blob/f36ce5a8b52db22e1ff1db46efb5209f344c3693/internal/api/middleware.go#L86

Use https://golang.org/pkg/net/#ParseCIDR on the hosts, if there is an error, just do the current Contains check, otherwise parse the host with https://golang.org/pkg/net/#ParseIP and use the https://golang.org/pkg/net/#IPNet.Contains method to check the IP.

Makefile
jstrachan: whoops! lemme re-do this PR with just the browser URL

internal/api/api.go
jstrachan: sounds good to me

jstrachan:
replacing this PR with https://github.com/vmware-tanzu/octant/pull/921

will ponder how to do the accept host PR later on

jstrachan:
incidentally using listen on `0.0.0.0:0` is another fix for the accept-hosts issue ;)

PR #921, opened 2020-05-07
fix: add CLI argument `browser-path`

jstrachan:

adds a CLI argument:

* `browser-path` so you can specify the initial context path to open the browser as you start Octant.

**Release note**:
```

This release adds a new CLI argument:

* `browser-path` so you can specify the initial context path to open the browser as you start Octant

e.g. to open octant on the Deployments view:

    octant --browser-path "/#/overview/namespace/default/workloads/deployments"

```

### DIFF
--- a/internal/commands/dash.go
+++ b/internal/commands/dash.go
@@ -77,6 +77,7 @@ func newOctantCmd(version string) *cobra.Command {
 					KubeConfig:             viper.GetString("kubeconfig"),
 					Namespace:              viper.GetString("namespace"),
 					FrontendURL:            viper.GetString("ui-url"),
+					BrowserPath:            viper.GetString("browser-path"),
 					Context:                viper.GetString("context"),
 					ClientQPS:              float32(viper.GetFloat64("client-qps")),
 					ClientBurst:            viper.GetInt("client-burst"),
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -0,0 +1,16 @@
+
```
wwitzel3, line 1: remove this file

wwitzel3 (changes requested):
Looks good, just need to remove that Makefile from the PR.

Makefile
jstrachan: doh - can't believe I added it back again ;) sorry about that...

jstrachan: @wwitzel3 I just pushed an updated PR without that pesky Makefile ;)

wwitzel3 (approved):
LGTM, I want one more set of eyes on it from @vmware-tanzu/octant-reviewers to makes sure there isn't something I'm not thinking about.

GuessWhoSamFoo (approved):

GuessWhoSamFoo:
Not a blocker for this PR, but what should happen if there is a conflict between the URL path and default namespace?

jstrachan:
> Not a blocker for this PR, but what should happen if there is a conflict between the URL path and default namespace?

Not sure I follow; you mean what if the URL path is for a namespace that doesn't exist? If so there's nothing stopping folks typing the same invalid URL into a browser address bar too right? Hopefully Octant can report if a namespace doesn't exist.

GuessWhoSamFoo:
Octant will notify users of an invalid namespace.

There's a namespace flag that assumes starting on the root path. In this case, the URL path will win. I wanted to point out there can be flags with conflicting user intent.

PR #926, opened 2020-05-07
Show deleted items with indicators

bryanl:

This change does the following:
* Shows items that are being deleted using an animation
* Adds an indicator for containing the object status in lists

This change touches lots of files since we don't have a good way to separate testing the operation list handlers from their contents.

**Release note**:
```
Add indicator to show object status and animation for objects when they are being deleted
```


wwitzel3:
I'm going to start taking a look at this now.

wwitzel3 (approved):

wwitzel3:
Pods being deleted continues to read "Running" during the delete process. When you view the Pod details, we provide the message that the pod is currently being deleted, but in the table view, it stays running. Which is not a bug in this PR. And directly ties in to Octant doing better when it comes to object status.

I wanted to note it because I think this PR will likely increase the priority of better object status so that there is better feedback for the user during the delete action.

PR #928, opened 2020-05-10
Add --namespace-list flag to accept a slice of namespaces

### LINKED ISSUE #911: Static namespace list
Our cluster does not provide namespace list operation.

Provide a static list of namespace like : `--namespaces [a,b,c,d]`

<3

**Environment:**

- Octant version (use `octant version`): 0.12.1
- Kubernetes version (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-26T06:17:09Z", GoVersion:"go1.14", Compiler:"gc", Platform:"darwin/amd64"}
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 10.15.4

GuessWhoSamFoo:
Adds a `--namespace-list` flag which accepts a slice of namespaces in the form of `x,y,z` and validates each via `Get` rather than `List` for users who have restricted clusters.

- Fixes #911

### DIFF
--- a/changelogs/unreleased/928-GuessWhoSamFoo
+++ b/changelogs/unreleased/928-GuessWhoSamFoo
@@ -1,1 +1,1 @@
-Added `--namespaces` flag to accept a slice of namespaces
+Added `--namespace-list` flag to accept a slice of namespaces
--- a/changelogs/unreleased/928-GuessWhoSamFoo
+++ b/changelogs/unreleased/928-GuessWhoSamFoo
@@ -1,0 +1,1 @@
+Added `--namespaces` flag to accept a slice of namespaces
--- a/internal/api/namespaces_manager.go
+++ b/internal/api/namespaces_manager.go
@@ -123,12 +123,26 @@ func NamespacesGenerator(_ context.Context, config NamespaceManagerConfig) ([]st
 		return nil, errors.Wrap(err, "retrieve namespaces client")
 	}
 
+	providedNamespaces := namespaceClient.ProvidedNamespaces()
+
 	names, err := namespaceClient.Names()
 	if err != nil {
 		initialNamespace := namespaceClient.InitialNamespace()
 		names = []string{initialNamespace}
 	}
 
+	for _, namespace := range providedNamespaces {
+		found := false
+		for _, foundNamespace := range names {
+			if namespace == foundNamespace {
+				found = true
+				break
+			}
+		}
+		if !found && namespaceClient.HasNamespace(namespace) {
+			names = append(names, namespace)
+		}
+	}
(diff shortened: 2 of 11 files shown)


wwitzel3:
I can't decide if we should just extend the `--namespace` flag to support invoking with a single value or a list or if we should add a new flag. I envisioned this extending `--namespace`

GuessWhoSamFoo:
The only benefit to setting this up as a separate flag is if a user wants to provide namespaces but not necessarily start octant on any of them. I don't think that merits a different flag - going to change this

wwitzel3:
Make sure we document that the first namespace becomes the initial/default namespace when Octant loads.

wwitzel3:
Ok, sorry @GuessWhoSamFoo but now I'm leaning back towards a separate flag. When I start to think about how to document this flag, there ends up being too many AND/ORs in it and it gets a little muddy. I think you had it right the first time, but maybe we call it `--namespace-list` ? To help disambiguate it from `--namespace`.

`--namespace` - sets the startup namespaces.
`--namespace-list` - explicitly sets the namespaces Octant will provide in the namespace selector.

GuessWhoSamFoo:
Rebased on master and changed flag name

changelogs/unreleased/928-GuessWhoSamFoo
```diff
@@ -0,0 +1 @@
+Added `--namespaces` flag to accept a slice of namespaces
```
wwitzel3, line 1: Update this changelog entry.

merlindorin:
Top! thx

tonyfalabella:
Can `--namespace-list` use wildcards?  If not is it something you can add?  Our company has many namespaces but for a given team all namespaces have a certain pattern.

I'd like to be able to use like this:
`octant --namespace-list foo*,bar*,xyz`

GuessWhoSamFoo:
That sounds like a reasonable request. Opened an issue for tracking

merlindorin:
@GuessWhoSamFoo I'm not sure if it's a good idea.

Under the hood, --namespace-list provide a static namespace list (no rights on kubeAPI namespace list). With a feature like a wildcard... we will have to list all namespaces available (needs kubeAPI namespace list rights) and then we will apply a filter.

It will be hard for a user with no rights on kubeAPI namespace list to understand why `--namespace-list foo-bar` works and why `--namespace-list foo-*` does not.

Maybe it can be something like `--namespace-filter` ?

PR #938, opened 2020-05-15
Set up storybook for text component

bryanl:

Adds storybook for previewing components


GuessWhoSamFoo:
If some of the more complex components like editor can render properly on story, we should use storybook over the angular subproject implementation so we don't have to build out features from scratch.

bryanl:
> If some of the more complex components like editor can render properly on story, we should use storybook over the angular subproject implementation so we don't have to build out features from scratch.

I need to research as I'm not anywhere near an expert, but this may be possible.

PR #940, opened 2020-05-18
Show loading screen when content is not ready

### LINKED ISSUE #506: Show a better loading screen.
We should show a loading screen while we wait for content response length to be greater than 0.

https://github.com/vmware-tanzu/octant/blob/d48f76ced7939d837c1995f6b43acebf3d54bd47/web/src/app/modules/overview/overview.component.ts#L148

nfarruggiagl:

Show a better loading screen
Loading

- Fixes #506 

**Release note**:
```
release-note
```

### DIFF
--- a/changelogs/unreleased/506-nfarruggiagl-mklanjsek
+++ b/changelogs/unreleased/506-nfarruggiagl-mklanjsek
@@ -1,0 +1,1 @@
+Show spinner when content is not ready
--- a/web/src/app/modules/shared/services/content/content.service.ts
+++ b/web/src/app/modules/shared/services/content/content.service.ts
@@ -14,6 +14,7 @@ import {
   LabelFilterService,
 } from '../label-filter/label-filter.service';
 import { NamespaceService } from '../namespace/namespace.service';
+import { LoadingService } from '../loading/loading.service';
 
 export const ContentUpdateMessage = 'event.octant.dev/content';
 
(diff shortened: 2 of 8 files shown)


GuessWhoSamFoo:
There is an existing loading component which is used for showing the spinner for a process happening on the backed. The changes here are for the time it takes to load all frontend components from a content response.

We also discussed during standup today about adding a delay before showing a the spinner. Maybe add a timer for 500ms? (this might require some tuning so that the spinner doesn't seem annoying)

mklanjsek (changes requested):
Both Sam and I looked into this and played around with it.  In general it works great!  The only small issue is when response from backend returns very quickly, for example if you do not have any resources like in my case RBAC - in that case spinner shows so briefly resulting in noticeable flicker.  We recommend adding the initial delay before showing spinner (maybe 500-1000ms), so the spinner only shows up for longer queries and prevents the flicker.

mklanjsek:
@GuessWhoSamFoo please update if I forgot something!

mklanjsek:
OK, I'll finish that.

mklanjsek:
Made following changes:
- Added logic so spinner does not show up if request takes < 650ms,
- In case spinner shows up, make sure it stays on for at least 1sec,
- Make spinner work when changing namespaces,

mklanjsek (approved):

wwitzel3 (approved):

PR #941, opened 2020-05-19
Run app in electron

bryanl:

This is the first pass at running Octant in Electron. It does not boot
Octant's dev front end and requires it to be running beforehand.

* `go run build.go build-electron-dev` builds the binary for your platform. This change was about laying the basics. It was not tested on Linux or Windows.


wwitzel3 (changes requested):
Can we add `go install github.com/asticode/go-astilectron-bundler/astilectron-bundler` to our `go run build.go go-install` command with this PR.

wwitzel3:
I can't test it on Windows or Linux, the executable builds, fails to launch on both platforms for me, no output, frontend running on 4200 in both cases. Not a blocker as this is experimental dev work at the moment.

bryanl:
> Can we add `go install github.com/asticode/go-astilectron-bundler/astilectron-bundler` to our `go run build.go go-install` command with this PR.

I chose not to do this because we don't need to do the download. Instead, what we could be doing instead is a `go mod download`, then using `go run`. It ensures we are running the expected version of the binary.

bryanl:
> I can't test it on Windows or Linux, the executable builds, fails to launch on both platforms for me, no output, frontend running on 4200 in both cases. Not a blocker as this is experimental dev work at the moment.

Windows may be failing due to a missing directory. I will make this work better in a cross-platform way. I'm trying to reproduce the Linux issue.

wwitzel3 (approved):

PR #958, opened 2020-05-28
Resource viewer initial prototype

### LINKED ISSUE #898: Start resource viewer research
Begin the process of identifying the needs and goals of the resource view. @bryanl can add some details to this as well. We will start to work with @rossfenrick.

Questions
- how is RV currently being used?
- what are the limitations of the current RV?
- how do we want to present object status within the RV?
- how do we want to enable information filtering within the RV?

mklanjsek:
Added the initial version of Resource Viewer prototype as a Storybook story. Still a long way to go, but adding this now should enable us to iterate faster and easier validate the concept. 

- Fixes #898

Run storybook with `npm run storybook` and it will show up as one of the stories.


mklanjsek:
Thanks for feedback, @bryanl!  We will certainly address that in next iteration.

PR #959, opened 2020-05-28
Check if running electron via user agent

GuessWhoSamFoo:

Running in binary currently prints `process is not defined` to console.

xref https://github.com/electron/electron/issues/2288#issuecomment-337858978


web/src/app/modules/shared/services/electron/electron.service.ts
```diff
@@ -15,9 +15,8 @@ export class ElectronService {
    * Returns true if electron is detected
    */
   isElectron(): boolean {
-    return (
```
wwitzel3, line 18:
```
    if (typeof process === 'undefined') {
      return false;
    }
```

PR #988, opened 2020-06-16
Add support for ALB ingress using annotations

### LINKED ISSUE #987: ALB Ingress Controller "use-annotation" backends

AWS ALB Ingress Controller allows you to reference an annotation as a Ingress Backend.ServiceName like in this example https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/tasks/ssl_redirect/

If you apply that yaml example and try navigating to the Ingress view the following error is thrown:

```
2020-06-12T23:47:41.177-0400	ERROR	api/content_manager.go:130	generate content	{"client-id": "9cea0b24-ad28-11ea-ba16-acde48001122", "err": "generate content: add row for object: get status for object: build status for ingress: get service from object store: get object from object store: services \"ssl-redirect\" not found", "content-path": "overview/namespace/default/discovery-and-load-balancing/ingresses"}
	/private/tmp/octant-20200604-13237-14eiloi/src/github.com/vmware-tanzu/octant/internal/api/content_manager.go:130
	/private/tmp/octant-20200604-13237-14eiloi/src/github.com/vmware-tanzu/octant/internal/api/poller.go:86
	/private/tmp/octant-20200604-13237-14eiloi/src/github.com/vmware-tanzu/octant/internal/api/poller.go:95
	/private/tmp/octant-20200604-13237-14eiloi/src/github.com/vmware-tanzu/octant/internal/api/content_manager.go:106
```

This also affects the "Resource Viewer" as if you view a service "Resource Viewer" it will try to parse the ingress that the service is attached to and will fail finding this service.

The service to be ignored and no errors found when viewing the Resource Viewer.

Am happy to create a PR for this, I can see that backends are looked up in these 2 places:

* https://github.com/vmware-tanzu/octant/blob/master/internal/objectstatus/ingress.go#L177
* https://github.com/vmware-tanzu/octant/blob/master/internal/queryer/queryer.go#L405

It could just skip the backend `if ServicePort == "use-annotation"`.

And/Or it could `return nil, nil` if the error is NotFound from the DynamicCache. However Octant will still mark the Ingress object status has having an error as it expects the backend to be there unless we implement the first option as well.

**Environment:**

- Octant version (use `octant version`): 0.13.1
- Kubernetes version (use `kubectl version`): v1.18.3
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 10.15

GuessWhoSamFoo:
~AWS ALB Ingress controller have references to a service through annotations. Rather than writing ALB-specific logic in Octant, we can have the resource viewer and object status tell the user that the service cannot be found and offload more detailed views via plugins.~

This PR checks validates the action in the annotations if the service port is `use-annotation`.

- Fixes #987 

### DIFF
--- a/internal/objectstatus/ingress.go
+++ b/internal/objectstatus/ingress.go
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	ingressNoBackendsDefined = "No backends defined. All traffic will be sent to the default backend configured for the ingress controller."
+	ingressNoBackendsDefined   = "No backends defined. All traffic will be sent to the default backend configured for the ingress controller."
+	ingressAlbActionAnnotation = "alb.ingress.kubernetes.io/actions."
 )
 
 func runIngressStatus(ctx context.Context, object runtime.Object, o store.Store) (ObjectStatus, error) {
@@ -74,6 +75,15 @@ func (is *ingressStatus) run(ctx context.Context) (ObjectStatus, error) {
 	}
 
 	for _, backend := range backends {
+		if backend.ServicePort.String() == "use-annotation" {
+			albAction := ingressAlbActionAnnotation + backend.ServiceName
+			if _, ok := ingress.Annotations[albAction]; !ok {
+				status.SetError()
+				status.AddDetailf("Backend refers to annotations %q which does't exist", albAction)
+			}
+			continue
+		}
+
 		key := store.Key{
 			Namespace:  ingress.Namespace,
 			APIVersion: "v1",
--- a/changelogs/unreleased/988-GuessWhoSamFoo
+++ b/changelogs/unreleased/988-GuessWhoSamFoo
@@ -1,0 +1,1 @@
+Added support for an ingress to have `use-annotation` ServicePort
--- a/internal/printer/ingress.go
+++ b/internal/printer/ingress.go
@@ -97,6 +97,20 @@ func (i *IngressConfiguration) Create(options Options) (*component.Summary, erro
 		sections.AddText("Default Backend", "Default is not configured")
 	}
 
+	for _, rule := range ingress.Spec.Rules {
+		if rule.IngressRuleValue.HTTP == nil {
+			continue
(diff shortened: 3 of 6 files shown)


wwitzel3:
Excellent, thank you @GuessWhoSamFoo

wwitzel3:
So taking https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/tasks/ssl_redirect/ as the example, we'd want the resource viewer to be green (assuming both `user-service` and `default-service` existed)

wwitzel3:
So taking https://kubernetes-sigs.github.io/aws-alb-ingress-controller/guide/tasks/ssl_redirect/ as the example, we'd want the resource viewer to be green (assuming both `user-service` and `default-service` existed)

wwitzel3:
What if we checked if a servicePort was an int? And if it is not an int, we can add that backend and it's service port under the Ingress config? This is a more general approach and won't require a plugin.

GuessWhoSamFoo:
The PR is updated specifically to handle the non-int case specifically for a serviceName to match the action in an annotation. Added code has no impact on other users and the ALB ingress controller is supported by a k8s sig so it should be fine being included in octant core until these cases are encountered more frequently.

internal/objectstatus/ingress.go
```diff
@@ -74,6 +75,15 @@ func (is *ingressStatus) run(ctx context.Context) (ObjectStatus, error) {
 	}
 
 	for _, backend := range backends {
+		if backend.ServicePort.String() == "use-annotation" {
+			albAction := ingressAlbActionAnnotation + backend.ServiceName
+			if _, ok := ingress.Annotations[albAction]; !ok {
+				status.SetError()
+				status.AddDetailf("Backend refers to annotations %q which does't exist", albAction)
```
wwitzel3, line 82: I think we should grab the information out of the annotations and write it to the config card.

PR #1005, opened 2020-06-24
Add Dependabot for front-end deps

scothis:
Adds front-end npm dependencies defined in `./web/package.json` under Dependabot management. PRs will be created when there is a new release published for each dependency.

Continues #990 

Based on the current output from `npm outdated` merging this PR will generate a lot of PRs to update individual npm dependencies.

**Release note**:
```
n/a
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -10,3 +10,8 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: npm
+  directory: "/web"
+  schedule:
+    interval: weekly
+  open-pull-requests-limit: 10


.github/dependabot.yml
```diff
@@ -10,3 +10,8 @@ updates:
   schedule:
     interval: daily
   open-pull-requests-limit: 10
+- package-ecosystem: npm
+  directory: "/web"
+  schedule:
+    interval: daily
```
scothis, line 16:
```suggestion
    interval: weekly
```

Will check for updates on Mondays at 5am UTC

mklanjsek:
We would certainly prefer to have client side updates batched in order to prevent noise and creating unnecessary versions of package.json and package-lock.json file. A couple of solutions were proposed in following slack thread:

https://kubernetes.slack.com/archives/CM37M9FCG/p1593102723185800

@scothis let's make sure that is addressed in near future.

scothis:
> let's make sure that is addressed in near future

By near future, do you mean once Dependabot adds support? If so 👍

PR #1007, opened 2020-06-25
Normalize collapsed nav flyouts

scothis:
The collapsed navigation bar has two different visual representations
for items today. For navigation items with children there are flyouts,
if there are no children a popover is used instead. The visual
difference is distracting.

Now a flyout is always used. If there are no children, it exists without
children. The flyout header text is now always a link in addition to the
existing icon link. For items with children, this was redundant with the
first child item. Now the 'Overview' item is contributed to the sub nav
automatically when the nav bar is expanded.

<table>
<tr>
<th>Before</th>
<th>After</th>
</tr>
<tr>
<td>Screen Shot 2020-06-25 at 5 26 26 PM</td>
<td>Screen Shot 2020-06-25 at 5 10 26 PM</td>
</tr>
<tr>
<td>Screen Shot 2020-06-25 at 5 26 24 PM</td>
<td>Screen Shot 2020-06-25 at 5 10 30 PM</td>
</tr>
<tr>
<td>Screen Shot 2020-06-25 at 5 11 35 PM</td>
<td>Screen Shot 2020-06-25 at 5 10 55 PM</td>
</tr>
</table>

**Release note**:
```
Normalize navigation bar flyouts when collapsed
```

### DIFF
--- a/changelogs/unreleased/1007-scothis
+++ b/changelogs/unreleased/1007-scothis
@@ -1,0 +1,1 @@
+Normalize navigation bar flyouts when collapsed
--- a/internal/modules/clusteroverview/clusteroverview.go
+++ b/internal/modules/clusteroverview/clusteroverview.go
@@ -249,7 +249,7 @@ func (co *ClusterOverview) Generators() []octant.Generator {
 
 func rbacEntries(ctx context.Context, prefix, namespace string, objectStore store.Store, _ bool) ([]navigation.Navigation, bool, error) {
 	neh := navigation.EntriesHelper{}
-	neh.Add("Overview", "", false)
+
 	neh.Add("Cluster Roles", "cluster-roles",
 		loading.IsObjectLoading(ctx, namespace, store.KeyFromGroupVersionKind(gvk.ClusterRole), objectStore))
 	neh.Add("Cluster Role Bindings", "cluster-role-bindings",
@@ -265,7 +265,7 @@ func rbacEntries(ctx context.Context, prefix, namespace string, objectStore stor
 
 func storageEntries(ctx context.Context, prefix, namespace string, objectStore store.Store, _ bool) ([]navigation.Navigation, bool, error) {
 	neh := navigation.EntriesHelper{}
-	neh.Add("Overview", "", false)
+
 	neh.Add("Persistent Volumes", "persistent-volumes",
 		loading.IsObjectLoading(ctx, namespace, store.KeyFromGroupVersionKind(gvk.PersistentVolume), objectStore))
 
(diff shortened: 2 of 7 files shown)


scothis:
@mklanjsek this is mostly the result of my playing around learning the codebase. I'm particularly unsure of the scss and would appreciate any pointers

mklanjsek:
Interesting idea with single item flyouts, @scothis !
But, have to say they look kind of wonky to me - combining menu header, button and tooltip feels too much visually and potentially confusing.  

Have to think more about that, it would be interesting what @rossfenrick  thinks!

scothis:
> combining menu header, button and tooltip feels too much visually and potentially confusing

Just to be clear, this change removes the tooltip. If you checkout, build and play with the change, it flows much more naturally than the current behavior.

To summarize the changes:
- the tooltip is replace with a flyout (every nav icon now has a flyout)
- the header in the flyout is now a link (the same link that the icon button already is)
- the "Overview" link in each flyout is removed (it was redundant with the icon button and header)
- no changes for the expanded nav

wwitzel3:
I like this experience better, but without the root entry, the navigation shows both the root menu entry AND the currently selected as active.

flyout-nav

scothis:
@wwitzel3 good catch. I added a [quick heuristic](https://github.com/vmware-tanzu/octant/pull/1007/commits/3d2f5c41a8dfd7a4976c647f7a9d435acd1a488f) to make this better, but we'll either need to refactor the information hierarchy or expose the setting into the server-side declaration of the nav item for a proper fix. In order to avoid multiple active nav items (or no active nav item) each top-level nav item should have a unique URL prefix, and the exact match setting should always be false.

wwitzel3:
@mklanjsek was there more we were waiting for on this? I've been using it and it feels pretty natural to me.

When/if this merges, @scothis please open an issue to update the server side navigation to properly capture the root information.

mklanjsek:
@wwitzel3 still waiting review from @rossfenrick

scothis:
@mklanjsek I'm not aware of any double selection issues after https://github.com/vmware-tanzu/octant/pull/1007/commits/3d2f5c41a8dfd7a4976c647f7a9d435acd1a488f, can you provide steps to reproduce?

mklanjsek:
> @mklanjsek I'm not aware of any double selection issues after [3d2f5c4](https://github.com/vmware-tanzu/octant/commit/3d2f5c41a8dfd7a4976c647f7a9d435acd1a488f), can you provide steps to reproduce?

I was talking about [this](https://github.com/vmware-tanzu/octant/pull/1007#issuecomment-650624309), please ignore if that has been addressed.

rossfenrick:
sorry for the delay, been on vacation. it looks fine to me, more consistent. one thing i noticed (this PR did not introduce this) is that in the light mode the active nav item is using a white background, which is really hard to even notice. the active background is much more noticeable in dark mode. we should think about using a light gray in light mode

scothis:
Squashed and rebased

@rossfenrick let's deal with the active color in a separate issue.

scothis:
Follow up items tracked in https://github.com/vmware-tanzu/octant/issues/1084

PR #1024, opened 2020-06-29
Bumped client side dependencies in preparation for switch to dependabot

mklanjsek:
In order to minimize the initial noise once we switch to dependabot (#1005), I would like to update everything to latest first.


GuessWhoSamFoo:
Better move quick before the deps are out of date again :trollface:

PR #1034, opened 2020-06-30
clean up old ci/cd mentions and plugin docs

### LINKED ISSUE #886: how to find the current namespace from a plugin?
The built in views automatically refresh as you switch namespaces (via the top right drop down). E.g. view the Workspaces -> Deployments; switch the namespace and you see a different list of deployments.

I'd like to replicate that behaviour inside a plugin view; by passing in the current namespace context to any query made in the plugin:

```go
	ctx := request.Context()
	client := request.DashboardClient()
        ns := TODO

	dl, err := client.List(ctx, store.Key{
		APIVersion: "apps/v1",
		Kind:       "Deployment",
		Namespace:  ns,
	})
```

At first I assumed there was no need to pass in a namespace as the `client.List()` would do that automatically; but it looks like you need to pass an explicit namespace in to filter by the current namespace otherwise all namespaces are searched. The issue is I don't see a way to find the current namespace from inside a plugin; though I see how to list them all.

I notice the current namespace is passed into the URL paths for the built in components; but couldn't see a way get the current namespace injected into the router paths either. Is there a trick I've missed?

There's a few ways this could be exposed. Maybe as a method on the DashboardClient interface; or maybe a property on the `service.Request` so that a plugin could encode the current namespace into the path in case a view needs it.

Another option is to specify `CurrentNamespace bool` on the `store.Key` struct so that the `DashboardClient` implementation knows only query with the current namespace in mind.

wwitzel3:

- Fixes up a typo in docs.go for actions.
- Replaces our reference tables with a link to our storybook.
- Removes mentions of old Drone CI/CD (replaced with actions)
- Adds action example to plugin docs.

- Fixes #886


site/docs/master/plugins/reference.md
```diff
@@ -1,311 +1,4 @@
...
 
-View component | Description | Example
----------------|-------------|---------
-[component.Text](#component-text) | Displays a string | ![newtext](newtext.png)
-[component.Timestamp](#component-timestamp) | Displays Unix time in a human-readable form. Hover to see full datetime. | ![timestamp](timestamp.png)
-[component.Table](#component-table) | Create a table with rows and columns | ![table](table.png)
-[component.Summary](#component-summary) | Create a summary containing a header and content | ![summary](summary.png)
-[component.Link](#component-link) | Create a link to an object | ![link](link.png)
-[component.Quadrant](#component-quadrant) | Create a quadrant consisting of fours strings and a title. | ![quadrant](quadrant.png)
-[component.Labels](#component-labels) | Create a label from a key-value pair of strings. | ![labels](labels.png)
```
GuessWhoSamFoo, line 11: The old assets can be cleaned up

GuessWhoSamFoo (approved):

PR #1093, opened 2020-07-10
Add port-forward button on Service summary page

### LINKED ISSUE #721: Add port forward functionality to Service summary page
Currently, Octant exposes the ability to start a port-forward to a pod. In addition to pods, kubectl can also port-forward to services using `kubectl port-forward svc/foo`.

Octant should support starting a port-forward from the Service summary page.

It might be worth exploring a UI enhancement where the Service ports are treated as first class citizens from a UI perspective. Perhaps adding a new card? That way each port can have its own "Start Port Forward" button.

ipsi:
Lets the user enable port forwarding from the services summary page, similar to how `kubectl` allows the user to specify a service when setting up port-forwarding.

This has a few quirks that might not be obvious:

* If the service maps several ports to a single target port (weird but possible), then _all_ of those ports will display the same port-forwarding information.
* Because we have to forward directly to a pod, when enabling port-forwarding via a service, that same port-forwarding will also be visible on the pod summary page too.
* To make it easier to handle named ports, we resolve the named port to an integer port when generating the service summary information. We also display that port on the service summary page. That said, in complex deployments, a single service could be linking multiple groups of pods, and that named port could map to multiple integer ports, so displaying the resolved port could be confusing?

- Fixes #721 

**Release note**:
```
Service summary page now includes a button to enable port-forwarding for all ports listed
Service summary page resolves named ports against containers and displays the port
```

### DIFF
--- a/internal/portforward/service.go
+++ b/internal/portforward/service.go
@@ -9,9 +9,6 @@ import (
 	"context"
 	"fmt"
 	"github.com/vmware-tanzu/octant/internal/util/kubernetes"
-	appsv1 "k8s.io/api/apps/v1"
-	apiEquality "k8s.io/apimachinery/pkg/api/equality"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sort"
 	"sync"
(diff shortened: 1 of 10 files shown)


internal/portforward/service.go
```diff
@@ -129,8 +132,6 @@ type Service struct {
 	state    States
 }
 
-var _ PortForwarder = (*Service)(nil)
```
wwitzel3, line 132: Add this back in. This is used a compile time check to ensure that the struct satisfies the interface.

internal/portforward/service.go
```diff
@@ -175,6 +176,15 @@ func (s *Service) validateCreateRequest(r CreateRequest) error {
...
 
+func (s *Service) FindPodSpecForService(ctx context.Context, apiVersion, kind, namespace, name string) (*corev1.PodSpec, error) {
+	pod, err := s.findPodForService(ctx, apiVersion, kind, namespace, name)
+	if err != nil {
+		return nil, err
+	}
+
+	return &pod.Spec, nil
+}
+
```
wwitzel3, line 187: It doesn't look like this is ever used inside or outside of the package, remove.

internal/portforward/service.go
ipsi: Spotted another one we'd removed, which appears to be the same pattern, `var _ error = (*notFound)(nil)` which I've added back in.

ipsi:
Changes all applied, should be good to go now.

wwitzel3:
Oh, a final thing, please add a changelog entry. See https://github.com/vmware-tanzu/octant/blob/master/CONTRIBUTING.md#changelog

PR #1094, opened 2020-07-10
Apply YAML UI

### LINKED ISSUE #954: Feature: Apply YAML
Have to use kubectl to apply new YAML

An editor window that I could copy and paste new YAML in to and apply it.

First pass implementation will be very simple.

- Provide a create link in the header
- Render an editor windows with an apply button
- Clicking apply will attempt to apply the YAML
- Namespace will be currently selected Octant namespace.

Errors:
- Errors will be displayed above the editor component?
- Errors will be sent as alerts?

**Environment:**

- Kubernetes version (use `kubectl version`):
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):

scothis:

Adds a link in the header to an empty YAML editor that uploads resources
to the cluster and defines an action `action.octant.dev/apply` which
parses the YAML and creates or updates each resources on the cluster.

Resources are processed and applied in the order they are defined.
New resources are created and existing resources are patched with the
changes applied server side, which requires k8s 1.16+.

Errors will stop the processing of new resources. Successfully applied
resources will be reported alongside any errors. When multiple resources
are applied, a summary is reported. Errors are lightly processed and may
not be user friendly.

Namespaced resources that do not define a namespace are defaulted to
the current namespace.

- Fixes #954

There are no tests at the moment, opening this PR to share the implementation and to gather feedback.

Screen Shot 2020-07-10 at 4 29 19 PM
Screen Shot 2020-07-10 at 4 25 00 PM
Screen Shot 2020-07-10 at 4 28 44 PM
Screen Shot 2020-07-10 at 4 25 31 PM
Screen Shot 2020-07-10 at 4 25 50 PM
Screen Shot 2020-07-10 at 4 26 30 PM

**Release note**:
```
Upload YAML applying to current cluster and namespace
```

### DIFF
--- a/internal/modules/configuration/applyyaml_describer.go
+++ b/internal/modules/configuration/applyyaml_describer.go
@@ -1,0 +1,46 @@
+/*
+Copyright (c) 2020 the Octant contributors. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package configuration
+
+import (
+	"context"
+
+	"github.com/vmware-tanzu/octant/internal/describer"
+	"github.com/vmware-tanzu/octant/internal/octant"
+	"github.com/vmware-tanzu/octant/pkg/view/component"
+)
+
+// ApplyYamlDescriber describes an apply
+type ApplyYamlDescriber struct {
+}
+
+var _ describer.Describer = (*ApplyYamlDescriber)(nil)
+
+// Describe describes the apply yaml interface
+func (d *ApplyYamlDescriber) Describe(ctx context.Context, namespace string, options describer.Options) (component.ContentResponse, error) {
+	title := append([]component.TitleComponent{}, component.NewText("Apply YAML"))
+	editor := component.NewEditor(component.TitleFromString("YAML"), "", false)
+	editor.Config.SubmitLabel = "Apply"
+	editor.Config.SubmitAction = octant.ActionApplyYaml
+	list := component.NewList(title, []component.Component{editor})
+
+	return component.ContentResponse{
+		Components: []component.Component{list},
+	}, nil
+}
+
+func (d *ApplyYamlDescriber) PathFilters() []describer.PathFilter {
+	filter := describer.NewPathFilter("/apply", d)
+	return []describer.PathFilter{*filter}
+}
+
+func (d *ApplyYamlDescriber) Reset(ctx context.Context) error {
+	return nil
+}
+
+func NewApplyYamlDescriber() *ApplyYamlDescriber {
+	return &ApplyYamlDescriber{}
+}
--- a/internal/octant/apply_yaml.go
+++ b/internal/octant/apply_yaml.go
@@ -1,0 +1,221 @@
+/*
+Copyright (c) 2020 the Octant contributors. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package octant
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/pkg/errors"
+	"github.com/vmware-tanzu/octant/internal/cluster"
+	"github.com/vmware-tanzu/octant/pkg/action"
+	"github.com/vmware-tanzu/octant/pkg/log"
+	"github.com/vmware-tanzu/octant/pkg/store"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/yaml"
+	sigyaml "sigs.k8s.io/yaml"
+)
+
+// ApplyYaml creates a yaml applier
+type ApplyYaml struct {
+	logger        log.Logger
+	objectStore   store.Store
+	clusterClient cluster.ClientInterface
+}
+
+var _ action.Dispatcher = (*ApplyYaml)(nil)
+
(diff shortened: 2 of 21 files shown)


wwitzel3:
Just confirming this is no longer a draft since you requested review?

internal/modules/configuration/uploads_describer.go
```diff
@@ -0,0 +1,46 @@
...
+	editor.Config.SubmitAction = octant.ActionApplyYaml
+	list := component.NewList(title, []component.Component{editor})
+
+	return component.ContentResponse{
+		Components: []component.Component{list},
+	}, nil
+}
+
+func (d *UploadsDescriber) PathFilters() []describer.PathFilter {
+	filter := describer.NewPathFilter("/upload", d)
```
wwitzel3, line 36: I think I'd prefer all of the references to be Apply to match convention. I can see the "upload YAML to apply it" but the way the UI functions, pasting it in to a box, that feels more like, Apply where a file picker would feel like Upload.

web/src/app/modules/sugarloaf/components/smart/container/container.component.html
```diff
@@ -15,6 +15,10 @@
       </div>
     </div>
     <div class="header-actions">
+      <a [routerLink]="['/configuration/upload']" class="header-link">
+        <clr-icon shape="upload"></clr-icon>
+        <span>Upload</span>
```
wwitzel3, line 20: Same here, lets call it `Apply YAML` and keep the upload arrow. I think that just the word upload is too vauge.

internal/octant/apply_yaml.go
```diff
@@ -0,0 +1,221 @@
...
+
+			return nil
+		}
+
+		// update object
+		unstructuredYaml, err := sigyaml.Marshal(doc)
+		if err != nil {
+			return fmt.Errorf("unable to marshal resource as yaml: %w", err)
+		}
+		True := true
```
wwitzel3, line 103: I understand why and even though we never actually change the value, I'd rather see this called `withForce` or even `Force` because then I at least get a hint at what this is doing. I'd like to see it moved to right before the if/else that is using it.

wwitzel3 (changes requested):
This is great. Nicely done. I have a couple minor nitpicks around the consistency of Upload vs. Apply and silly preference on a bool variable.

I like the Apply button being in the header vs. under the plugin nav. And I like the upload icon choice.

internal/modules/configuration/uploads_describer.go
scothis: I started off with "Upload" because their maybe other mechanism besides "apply". But since that's all we have, I agree "apply" makes more sense.

scothis:
> Just confirming this is no longer a draft since you requested review? Edit: I'll just assume yes.

It's a draft because I haven't written any tests yet, but otherwise it's ready

mklanjsek (dismissed):
The Apply YAML functionality looks great and works good in my testing! 

However, after playing with it for a while, in my opinion we should name it `New Resource` instead. From the end user point of view, that's what they'll be doing - creating new resources.
 
This is a completely new and ground breaking functionality in Octant and I think we need to think carefully about naming and positioning it in the UI. It also might be worth discussing that on community meeting.

wwitzel3:
> The Apply YAML functionality looks great and works good in my testing!
> 
> However, after playing with it for a while, in my opinion we should name it `New Resource` instead. From the end user point of view, that's what they'll be doing - creating new resources.
> 
> This is a completely new and ground breaking functionality in Octant and I think we need to think carefully about naming and positioning it in the UI. It also might be worth discussing that on community meeting.

I'm actually fine with the button in the header reading something like Add/Create/New Resource .. most people if they were updating resources via Octant probably are using the YAML tab for the resource. Which is the only argument for keeping it "apply" since that can mean both things.

I do want to keep the apply wording in the action and on the backend. But I don't have a strong opinion on the button name, except as mentioned, I didn't care for the term Upload.

scothis:
> we should name it `New Resource` instead. From the end user point of view, that's what they'll be doing - creating new resources.

Sometimes yes, sometimes not. `kubectl apply`, the most analogous comparison, is distinct from `kubectl create`. `apply` can create new resource, or it can patch an existing resource. While `create` requires the resource to not already exist.

> most people if they were updating resources via Octant probably are using the YAML tab for the resource. Which is the only argument for keeping it "apply" since that can mean both things.

Other significant differences include:
- any resource can be updated without first needing to find it in the UI (including resources not otherwise exposed
- multiple resources can be applied in the same request

wwitzel3:
> * any resource can be updated without first needing to find it in the UI (including resources not otherwise exposed)

To me, this is the big one. I do think even first time users will know the `apply` convention. And by keeping it `Apply YAML` we enable more advanced users a quick short cut.

What about a tooltip? `Create or update existing resources`

scothis:
> What about a tooltip? 

Added `Create or update resources`

internal/modules/configuration/applyyaml_describer.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/vmware-tanzu/octant/pkg/view/component"
+)
+
+// ApplyYamlDescriber describes an upload
+type ApplyYamlDescriber struct {
+}
+
+var _ describer.Describer = (*ApplyYamlDescriber)(nil)
+
+// Describe describes the upload yaml interface
```
wwitzel3, line 22: Missed the comments.

wwitzel3 (approved):
LGTM, minor nit about the comments still saying `upload` that should be fixed up before merge, but not a blocker.

internal/modules/configuration/applyyaml_describer.go
scothis: All non-icon "uploads" are now gone

PR #1096, opened 2020-07-10
Fixed storybook build

mklanjsek:
Storybook build got broken after update to @types/node in #1070.  This change reverts it back to original version `12.12.40`


mklanjsek:
Closing, Git got out of sync here, will continue with #1097

Issue #1084, opened 2020-07-08
Add contrast to active navigation items


scothis:
**Describe the problem/challenge you have**

In the light theme, the active background color for navigation bar items is white and difficult to see against the very light grey nav background. The dark theme has much more contrast.

<table>
<tr>
<th>Light</th>
<th>Dark</th>
</tr>
<tr>
<td>
<img width="65" alt="Screen Shot 2020-07-08 at 5 38 13 PM" src="https://user-images.githubusercontent.com/302992/86973137-1571c000-c142-11ea-8542-8d6ba5a3f699.png">
</td>
<td>
<img width="70" alt="Screen Shot 2020-07-08 at 5 38 29 PM" src="https://user-images.githubusercontent.com/302992/86973147-1a367400-c142-11ea-83b8-9908e44db7b3.png">
</td>
</tr>
</table>

**Describe the solution you'd like**

Minimally the active background color needs to have more contrast in light mode.

**Anything else you would like to add:**

Follow up from https://github.com/vmware-tanzu/octant/pull/1007#issuecomment-655675676

GuessWhoSamFoo:
cc @rossfenrick

PR #1100, opened 2020-07-13
Fix hover on collapsed navigation

GuessWhoSamFoo:
Fixes:

hover


mklanjsek (changes requested):
It looks good, except that it doesn't seem to work with the light theme.

mklanjsek:
The light theme behavior could be related to #1084

wwitzel3:
dismiss-flyout

Works well for me in both dark and light mode.

wwitzel3 (approved):

mklanjsek:
The icon is not changing color on hover in light theme the same way it's doing in the dark theme. You can see that in the above recording from Wayne. It's probably just not the correct color applied in light theme.

wwitzel3:
Yeah, that's covered by #1084 that's not a blocker for this.

image

mklanjsek (approved):
We can handle light theme color in separate PR

PR #1133, opened 2020-07-18
use options.KubeConfig to check if loading API is needed

### LINKED ISSUE #1128: kubeconfig flag not working in 0.14
Use the `--kubeconfig` should set the config even if there is not a default one present.
`--kubeconfig` should take precedence over the default config if it is present.

The ENVAR seems to work still.

**Environment:**

- Octant version (use `octant version`):  0.14
- Kubernetes version (use `kubectl version`): 0.16
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Windows/Ubuntu

### LINKED ISSUE #1129: prompted to enter kubeconfig even though I provided one
I have provided a kubeconfig via `KUBECONFIG`. Octant starts and the logging shows it is using my config and creating the API service.

When the browser opens, I am still prompted for a kubeconfig.
image (13)

Version: 0.14
OS: Ubuntu/Chrome

wwitzel3:

- Fixes #1128 
- Fixes #1129

### DIFF
--- a/changelogs/unreleased/1129-wwitzel3
+++ b/changelogs/unreleased/1129-wwitzel3
@@ -1,0 +1,2 @@
+Fixed bug where the client would request from the loading API with a valid kubeconfig.
+Fixed bug where --kubeconfig flag would not work.
--- a/pkg/dash/dash.go
+++ b/pkg/dash/dash.go
@@ -20,7 +20,6 @@ import (
 	"github.com/soheilhy/cmux"
 	"github.com/spf13/viper"
 	"go.opencensus.io/trace"
-	"k8s.io/client-go/tools/clientcmd"
 
 	"github.com/vmware-tanzu/octant/internal/api"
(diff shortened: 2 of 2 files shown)


wwitzel3:
> Tested on my machine and it works well (note: I was also not able to repro the issue on my machine before the fix).
> 
> Notice that when I remove `current-context` field from bubeconfig file, Octant fails to start with this:
> 
> ```
> unable to start runner: failed to start service api: failed to init cluster client: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable
> ```
> 
> Since that used to make Octant show the `Add a new kubeconfig` popup, I thing this is much better than before (and as Wayne just pointed out to me, `kubectl` fails in that case too).

I extended our error message to hint at checking the current-context

GuessWhoSamFoo:
Testing various ways to provide a kubeconfig (command line flag, environment variable, manual input through UI) will be need be expanded upon from the e2e proposal in #810

PR #1144, opened 2020-07-22
Display error when unable to find valid local port

### LINKED ISSUE #497: Alert user when unable to find a local port
When creating a port forward if unable to find an open local port Octant should alert the user.

https://github.com/vmware-tanzu/octant/blob/db1afb6703d42db632c5903ed8990f9946b55a5f/internal/portforward/portforward.go#L89

ipsi:
Displays an error when port-forwarding cannot occur due to an error (e.g. unable to resolve a local port)

- Fixes #497

Passing the `Alerter` all the way down the call chain is not ideal, but the other options have their own problems:

1. Blocking the `Handle` function until port-forwarding setup is complete
  This would work, but it doesn't seem to make a lot of sense given that this is currently done in a go-func, and the `Handle` function is called synchronously by the web socket process. If we want to block on the handle method, it might make more sense to not use go-funcs at all here and have the entire chain become blocking, and just pass the `err` back up in the normal way.
  It's not clear what other implications making this synchronous would have, so it feels like quite a big change.
1. Moving the `localPortsHandler` function up a few levels, so the `Handle` function can call it directly and pass in the `Alerter`
  This isn't really an option either - the `localPortsHandler` function relies on an `Options` struct that's created and passed in from `Service#createForwarder`, and passing that back up the chain doesn't seem appropriate.
1. Modify the `DefaultPortForwarder` struct to hold an `Alerter`
  

**Release note**:
```
Display alert if there is an error setting up port-forwarding
```

### DIFF
--- a/internal/portforward/portforward.go
+++ b/internal/portforward/portforward.go
@@ -6,14 +6,14 @@ SPDX-License-Identifier: Apache-2.0
 package portforward
 
 import (
+	"github.com/vmware-tanzu/octant/pkg/action"
 	"io"
-	"net/http"
-	"net/url"
-
 	"k8s.io/client-go/rest"
 	restclient "k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/portforward"
 	"k8s.io/client-go/transport/spdy"
+	"net/http"
+	"net/url"
 )
 
 // Options contains all the options for running a port-forward
@@ -30,7 +30,7 @@ type Options struct {
 }
 
 type portForwarder interface {
-	ForwardPorts(method string, url *url.URL, opts Options) error
+	ForwardPorts(alerter *action.Alerter, method string, url *url.URL, opts Options) error
 }
 
 type DefaultPortForwarder struct {
@@ -53,7 +53,7 @@ type ForwardedPort struct {
 	Remote uint16
 }
 
-func (f *DefaultPortForwarder) ForwardPorts(method string, url *url.URL, opts Options) error {
+func (f *DefaultPortForwarder) ForwardPorts(alerter *action.Alerter, method string, url *url.URL, opts Options) error {
 	transport, upgrader, err := spdy.RoundTripperFor(opts.Config)
 	if err != nil {
 		return err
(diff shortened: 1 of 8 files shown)


ipsi:
Squashed the four misc commits into one.

internal/portforward/portforward.go
```diff
@@ -30,7 +32,7 @@ type Options struct {
 }
 
 type portForwarder interface {
-	ForwardPorts(method string, url *url.URL, opts Options) error
+	ForwardPorts(alerter *action.Alerter, method string, url *url.URL, opts Options) error
```
wwitzel3, line 35: No `*` here.

internal/portforward/portforward.go
```diff
@@ -53,7 +55,7 @@ type ForwardedPort struct {
 	Remote uint16
 }
 
-func (f *DefaultPortForwarder) ForwardPorts(method string, url *url.URL, opts Options) error {
+func (f *DefaultPortForwarder) ForwardPorts(alerter *action.Alerter, method string, url *url.URL, opts Options) error {
```
wwitzel3, line 58: Same, no `*` as `action.Alerter` is an interface.

internal/portforward/portforward.go
```diff
@@ -84,21 +86,27 @@ func localPortsHandler(fw *portforward.PortForwarder, opts Options) {
 		go func() {
 			select {
 			case <-opts.ReadyChannel:
-				ports, err := fw.GetPorts()
+				forwardedPorts, err := fw.GetPorts()
 				if err != nil {
-					// TODO: alert user (GH#497)
+					(*alerter).SendAlert(action.CreateAlert(action.AlertTypeError, "Error resolving local port: "+err.Error(), action.DefaultAlertExpiration))
```
wwitzel3, line 91: No need for this once you convert the argument to just taking the interface not a pointer to the interface.
You probably want to check if `alerter` is nil to avoid a panic output that could happen if this go routine was finishing up when Octant was interrupted / killed.

Issue #1136, opened 2020-07-20
Allow plugins to apply yaml


wwitzel3:
Now that #1094 is merged.

Extend this functionality to plugins.

wwitzel3:
This was added for the JavaScript plugin runtime in #1141

PR #1152, opened 2020-07-23
Refactor apply_yaml action to use ObjectStore.CreateFromYAML

wwitzel3:

Make it easier for other parts of Octant to take advantage of the CreateFromYAML logic that was added in #954

### DIFF
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -38,6 +38,7 @@ type Store interface {
 	Update(ctx context.Context, key Key, updater func(*unstructured.Unstructured) error) error
 	IsLoading(ctx context.Context, key Key) bool
 	Create(ctx context.Context, object *unstructured.Unstructured) error
+	CreateOrUpdateFromYAML(ctx context.Context, namespace, input string) ([]string, error)
 }
 
 // Key is a key for the object store.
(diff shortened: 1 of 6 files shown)


wwitzel3:
related #1136

pkg/store/store.go
```diff
@@ -38,6 +38,7 @@ type Store interface {
 	Update(ctx context.Context, key Key, updater func(*unstructured.Unstructured) error) error
 	IsLoading(ctx context.Context, key Key) bool
 	Create(ctx context.Context, object *unstructured.Unstructured) error
+	CreateFromYAML(ctx context.Context, namespace, input string) ([]string, error)
```
scothis, line 41: "Create" is only half of what this call actually does.

```suggestion
	ApplyFromYAML(ctx context.Context, namespace, input string) ([]string, error)
```

or 

```suggestion
	CreateOrUpdateFromYAML(ctx context.Context, namespace, input string) ([]string, error)
```

scothis (approved):
Looks good. One small quibble with the method name

pkg/store/store.go
wwitzel3: Yep, this is better. I'll change it.

scothis (approved):

PR #1175, opened 2020-07-28
Switch Mutating/ValidatingWebhookConfiguration version to v1beta1

### LINKED ISSUE #1168: ValidatingWebhookConfiguration v1beta1 conversion v1 error.
We should do our best to degrade down to v1beta1 from v1 in this case. Lots of common tools still install v1beta1.

Resource Viewer breaks with
```
create resource viewer: unable to visit /v1, Kind=Service jx-golang-http-test: error unable to visit object </v1, Kind=Service jx-golang-http-test>: converting unstructured validatingwebhookconfiguration: scheme convert: converting (v1beta1.ValidatingWebhookConfiguration) to (v1.ValidatingWebhookConfiguration): unknown conversion
```

Console output when attempting to visit /cluster-overview/api-server/validating-webhooks
```
2020-07-27T06:17:15.569-0400    ERROR   api/content_manager.go:130      generate content        {"client-id": "01341e68-cff1-11ea-a7f0-7617a1e4e1a6", "err": "generate content: scheme convert: converting (v1beta1.ValidatingWebhookConfiguration) to (v1.ValidatingWebhookConfiguration): unknown conversion", "content-path": "cluster-overview/api-server/validating-webhooks"}
        /home/wwitzel3/repos/octant/internal/api/content_manager.go:130
        /home/wwitzel3/repos/octant/internal/api/poller.go:86
        /home/wwitzel3/repos/octant/internal/api/poller.go:95
        /home/wwitzel3/repos/octant/internal/api/content_manager.go:106
```

scothis:

Cluster before Kubernetes 1.16 don't support admissionregistration.k8s.io/v1.
The v1beta1 version was deprecated in k8s 1.16, but is still available
and semantically equivalent.

- Fixes #1168 

This change also fixes a bug in the ObjectStore where it might return
objects for a different version than requested. Instead of the requested
version, the server's default version for the resource was used. With
this change, it is possible that we will have multiple informers for the
same resource at different versions. While inefficient the conversion
logic is on the server, where it belongs.

**Release note**:
```
none
```

### DIFF
--- a/internal/objectstore/informer.go
+++ b/internal/objectstore/informer.go
@@ -71,6 +71,7 @@ func (f *informerFactory) ForResource(groupVersionKind schema.GroupVersionKind) 
 	stopCh := f.informerContextCache.addChild(groupVersionKind)
 
 	gvr, _, err := f.client.Resource(groupVersionKind.GroupKind())
+	gvr.Version = groupVersionKind.Version
 	if err != nil {
 		return nil, fmt.Errorf("unable to find group version resource for group kind %s: %w",
 			groupVersionKind.GroupKind(), err)
--- a/internal/cluster/cluster.go
+++ b/internal/cluster/cluster.go
@@ -14,7 +14,7 @@ import (
 	"time"
 
 	"github.com/pkg/errors"
-	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
+	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
 	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/install"
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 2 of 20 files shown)


internal/objectstore/informer.go
```diff
@@ -71,6 +71,7 @@ func (f *informerFactory) ForResource(groupVersionKind schema.GroupVersionKind)
 	stopCh := f.informerContextCache.addChild(groupVersionKind)
 
 	gvr, _, err := f.client.Resource(groupVersionKind.GroupKind())
+	gvr.Version = groupVersionKind.Version
```
scothis, line 74: This is actually the root cause of the error. With only this change on a cluster that doesn't support admissionregistration.k8s.io/v1 you wouldn't see a conversion error, but an empty listing.

Issue #1102, opened 2020-07-13
Add Stepper Component


wwitzel3:
Add Stepper Component to Octant
 - https://clarity.design/documentation/stepper

Deliverable
 - Storybook entry that shows now this component will work with some mock data.
 - An angular component.
 - A Go component with a marshal/unmarshal implementation.

No server state, this is just the component.

wwitzel3:
Storybook and Angular component are completed.

Left TODO:
 - [ ] add the component to the content switcher ngFor
 - [ ] implement the Go Marshal/Unmarshal.

wwitzel3:
@nanaasiedu pointed out that the Props are not rendering, this is something we will need to address as well.

wwitzel3:
I'm struggling with Storybook and the component in general. Lots of console errors around the formBuilder. I am thinking I will revisit this using form templates instead? Maybe that will be a more friendly way in Angular to render a number of undetermined form elements from a config.

wwitzel3:
I played with this some more yesterday, didn't make any direct headway, but I didn't find another project wanting to do something similar, essentially rendering form inputs on the fly.

I think, the trick is to stub out the formGroup sections with empty controls so it matches anything and then as we add elements we an add the control so it can match / validate.

wwitzel3:
I managed to get the formGroup working by ensuring that config and items were both set before attempting to forLoop.
The only remaining issue is Props and Actions/Knobs on the story book pages are not displaying.

wwitzel3:
I found why props were not displaying. Needed to be added to the compudoc tsconfig.
Final thing is figuring out what is causing the undefined error when using the Canvas.

```
vendors~main.87fc3cbf4a94149f9543.bundle.js:128046 ERROR TypeError: Cannot read property 'id' of undefined
    at StepperModel.push../node_modules/@clr/angular/fesm5/clr-angular.js.StepperModel.openFirstPanel (vendors~main.87fc3cbf4a94149f9543.bundle.js:124133)
    at StepperModel.push../node_modules/@clr/angular/fesm5/clr-angular.js.StepperModel.updatePanelOrder (vendors~main.87fc3cbf4a94149f9543.bundle.js:124075)
    at StepperService.push../node_modules/@clr/angular/fesm5/clr-angular.js.AccordionService.updatePanelOrder (vendors~main.87fc3cbf4a94149f9543.bundle.js:123622)
[...]

PR #1177, opened 2020-07-29
Add stepper component.

wwitzel3:

image

image

### DIFF
--- a/web/src/app/modules/shared/components/presentation/stepper/stepper.component.ts
+++ b/web/src/app/modules/shared/components/presentation/stepper/stepper.component.ts
@@ -9,7 +9,12 @@ import {
   OnDestroy,
   isDevMode,
 } from '@angular/core';
-import { StepperView, View, ActionField } from '../../../models/content';
+import {
+  StepperView,
+  View,
+  ActionField,
+  StepItem,
+} from '../../../models/content';
 import {
   FormGroup,
   FormBuilder,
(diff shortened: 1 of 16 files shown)


wwitzel3:
@GuessWhoSamFoo or @mklanjsek can either of you take a look at this and see if you can figure out what is causing the property error?

mklanjsek:
Started looking at this right now

wwitzel3:
Thanks @mklanjsek , the error I am seeing is in the issue, https://github.com/vmware-tanzu/octant/issues/1102#issuecomment-666341976

I only see it the first time I attempt to load the component in the Cnavas, this error does not occur when it first loads in to the docs.

wwitzel3:
I'll rebase after review.

web/src/app/modules/shared/components/presentation/stepper/stepper.component.ts
```diff
@@ -0,0 +1,118 @@
...
+  steps: StepItem[];
+
+  constructor(
+    private formBuilder: FormBuilder,
+    private websocketService: WebsocketService
+  ) {
+    this.steps = [] as StepItem[];
+  }
+
+  ngOnChanges(changes: SimpleChanges): void {}
```
mklanjsek, line 64: We should handle stepper definition changes and update the component.  Without that the `view` knob in Storybook will never work.
(if you feel the stepper definition will remain static, than maybe just remove the knob)

wwitzel3: I'll remove it for now, we decided the initial implementation of the stepper would be static.

wwitzel3:
@mklanjsek PTAL

PR #1213, opened 2020-08-07
Create prototype for generating ts from components

bryanl:

To support typescript plugins, it should be easier to access components definitions in typescript without writing your library. This change is the start of a new library that will be generated from the go components. In the future, this could also be used by the frontend, so there is only one source of truth for components.

High level:
* Annotate component with `octant:component` tag
* Run generator

This command converts components to typescript.

requirements:
- prettier: `npm i -g prettier`

running:
```sh
$ go run ./cmd/ts-component-gen/main.go -dest <directory>
```
process:
- Figure out what components exist. This process relies on convention,
so components are labeled with an `+octant:component` tag in a doc
comment.
- Extract type information by building up a new binary that contains
reflect type information.
- Convert types to typescript
- Write typescript components to the supplied destination


wwitzel3:
With

```
    const t = new TextFactory({ value: "card body" });
    const card = new CardFactory({body: t});
```

body is `Component<any>` so I would expect this to work.

I get the following TypeScript error:
image

wwitzel3:
Never mind, totally missing `toComponent`

bryanl:
> Never mind, totally missing `toComponent`

Yeah. Still need to make this more obvious. The idea here was to separate the component factory from the actual component itself. We'll get better at this as we use it more.

bryanl:
🥳

PR #1216, opened 2020-08-08
Add SendAlert to plugin interface

### LINKED ISSUE #902: is there a way for a plugin to send an Alert event to the UI
plugins can provide actions via the `Button` / `Form` / `Action` mechanism which lets you invoke handlers inside the plugin which is awesome. It would be great to be able to return an Alert to the front end, similar to when you trigger a CronJob or delete a resource or whatever to show if the action completed or failed

Maybe the `service.ActionRequest` struct could also have an Alerter interface which lets you pass in an alert struct (text/markdown component and duration?)

GuessWhoSamFoo:
This PR allows plugin authors to send an alert via plugins. It also creates a new `pkg/event` package that can be used to expose Octant events in the future.

Sending an alert from a plugin will require a websocket client ID which is passed through `request.Context()`.

- Fixes #902


wwitzel3:
Does the state not have access to the client ID it was created for? Maybe we can add that to the state and then we don't need to pass the client all the way down just to get the ID

GuessWhoSamFoo:
Rebased and added a `GetClientID` in state.go to avoid needing to pass around the client in `content_manager.go`

PR #1217, opened 2020-08-08
restructure JavaScript plugin code

wwitzel3:

This restructures the existing JavaScript plugin runtime code making it easier to test and reason about. More can be done, but this is a good first pass.

Also added some basic sanity tests for the dashClient that the JavaScript runtime uses.

### DIFF
--- a/pkg/plugin/javascript/dashclient_test.go
+++ b/pkg/plugin/javascript/dashclient_test.go
@@ -1,0 +1,104 @@
+package javascript
+
+import (
+	"context"
+	"testing"
+
+	"github.com/dop251/goja"
+	"github.com/golang/mock/gomock"
+	"github.com/stretchr/testify/assert"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/vmware-tanzu/octant/pkg/store"
+	storeFake "github.com/vmware-tanzu/octant/pkg/store/fake"
+)
+
+func TestDashboardClient_Get(t *testing.T) {
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	ctx := context.Background()
+	objectStore := storeFake.NewMockStore(controller)
+
+	vm := goja.New()
+	vm.SetFieldNameMapper(goja.TagFieldNameMapper("json", true))
+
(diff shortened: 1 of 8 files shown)


pkg/plugin/javascript/dashclient_test.go
```diff
@@ -0,0 +1,100 @@
...
+	vm := goja.New()
+	vm.SetFieldNameMapper(goja.TagFieldNameMapper("json", true))
+
+	obj := CreateDashClientObject(ctx, objectStore, vm)
+	assert.NotNil(t, obj)
+
+	objectStore.EXPECT().Get(ctx, store.Key{APIVersion: "v1", Kind: "Pod"}).Return(&unstructured.Unstructured{}, nil)
+
+	vm.Set("dashClient", obj)
+	vm.RunString(`
```
bryanl, line 32: Check the return code of this item call to `RunString` and the others below.

pkg/plugin/javascript/dashclient_test.go
```diff
@@ -0,0 +1,100 @@
...
+dashClient.List({namespace:'test', apiVersion: 'v1', kind:'Pod'})
+`)
+}
+
+func TestDashboardClient_CreateUpdate(t *testing.T) {
+	controller := gomock.NewController(t)
+	defer controller.Finish()
+
+	ctx := context.TODO()
```
bryanl, line 62: This is a nit: context.TODO() is for when you are going to plumb a context in later. Use context.Background() instead. They are literally the same thing, but only differ intent.

pkg/plugin/javascript/httpclient.go
```diff
@@ -0,0 +1,102 @@
...
+	if err := client.Set("getJSON", h.getJSON); err != nil {
+		return vm.NewTypeError(fmt.Errorf("httpClient.Set.getJSON: %w", err))
+	}
+	if err := client.Set("post", h.post); err != nil {
+		return vm.NewTypeError(fmt.Errorf("httpClient.Set.post: %w", err))
+	}
+	return client
+}
+
+func (h *httpClient) _get(c goja.FunctionCall) (goja.Callable, []byte, error) {
```
bryanl, line 42: Why is this prefixed with an underscore?

PR #1219, opened 2020-08-09
Fix for UI stutter when changing views

### LINKED ISSUE #1205: UI stutter from old view to new view.
Since we added the loading indicator there is a very bad stutter in the UI that results in seeing the previously rendered page, long enough to think there has been a problem, before the new view renders.

See gif for example. At 17 seconds (bottom right corner) you I click back to the Overview. A full 3 seconds tick off with the load view being displayed AFTER the loading indicator was displayed.

ui-stutter

mklanjsek:
We now ignore premature UI updates that were happening before the frontend had data it needed for displaying the new view. This change also improves how we handle view changes between different namespaces and tweaks slightly the spinner timings.  
 
- Fixes #1205

### DIFF
--- a/changelogs/unreleased/1205-mklanjsek
+++ b/changelogs/unreleased/1205-mklanjsek
@@ -1,0 +1,1 @@
+Fixed UI stutter when changing views
--- a/web/src/app/modules/shared/models/content.ts
+++ b/web/src/app/modules/shared/models/content.ts
@@ -4,6 +4,7 @@
 
 export interface ContentResponse {
   content: Content;
+  currentPath: string;
 }
 
 export interface PathItem {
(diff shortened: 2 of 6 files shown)


GuessWhoSamFoo:
There is a case with clarity tabs switching back to the first tab on URL changes that causes a flicker. The new loading time looks fine locally - I'm not sure if it is too aggressive for larger clusters.

mklanjsek:
> There is a case with clarity tabs switching back to the first tab on URL changes that causes a flicker. The new loading time looks fine locally - I'm not sure if it is too aggressive for larger clusters.

The flicker happens only when you go back to the home page (e.g. click on Octant icon) and unfortunately we can’t do much about it because of the way routing and clarity are working in that case.

As for the timing - I tested with fairly large clusters and it was performing pretty good on my machine in both Firefox and Chrome - let me know if you find one that does not perform well.

GuessWhoSamFoo:
lgtm, please squash the changelog commit

PR #1231, opened 2020-08-10
fix: bump grpc max receive message size

### LINKED ISSUE #924: Plugin gRPC can exceed the default recv buffer of 4MB
Currently we use the default gRPC receive size of 4MB. We should considering increasing this.

See https://github.com/bloodorangeio/octant-helm/issues/4
See https://kubernetes.slack.com/archives/CM37M9FCG/p1588858729090800

nodece:
Signed-off-by: Zixuan Liu <nodeces@gmail.com>

This change is increase grpc max receive message size (4MB to 16 MB), and support set grpc max receive message size from `OCTANT_CLIENT_MAX_RECV_MSG_SIZE` environment variables.

- Fixes https://github.com/vmware-tanzu/octant/issues/924

**Release note**:
```
Increase grpc max receive message size (4MB to 16 MB), and support set grpc max receive message size from `OCTANT_CLIENT_MAX_RECV_MSG_SIZE` environment variables
```

### DIFF
--- a/internal/commands/dash.go
+++ b/internal/commands/dash.go
@@ -154,6 +154,7 @@ func newOctantCmd(version string, gitCommit string, buildTime string) *cobra.Com
 	octantCmd.Flags().StringSlice("namespace-list", []string{}, "a list of namespaces to use on start")
 	octantCmd.Flags().StringP("plugin-path", "", "", "plugin path")
 	octantCmd.Flags().BoolP("verbose", "v", false, "turn on debug logging")
+	octantCmd.Flags().IntP("client-max-recv-msg-size", "", 1024*1024*16, "client max receiver message size")
 
 	octantCmd.Flags().StringP("accepted-hosts", "", "", "accepted hosts list [DEV]")
 	octantCmd.Flags().Float32P("client-qps", "", 200, "maximum QPS for client [DEV]")
--- a/pkg/plugin/api/api.go
+++ b/pkg/plugin/api/api.go
@@ -9,6 +9,8 @@ import (
 	"context"
 	"net"
 
+	"github.com/spf13/viper"
+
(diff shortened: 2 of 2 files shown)


wwitzel3:
Thank you for this PR.

We use the cobra and viper packages to expose CLI and ENV variables in Octant. Could you update this to follow the convention of our other runtime options.

See: https://github.com/vmware-tanzu/octant/blob/bd4105ffb92602f62ae227c9788e85bf9b86b484/internal/commands/dash.go#L145 and https://github.com/vmware-tanzu/octant/blob/bd4105ffb92602f62ae227c9788e85bf9b86b484/internal/commands/dash.go#L83

nodece:
@wwitzel3 good idea. let me try!

nodece:
@wwitzel3 Done!

internal/commands/dash.go
```diff
@@ -191,6 +192,9 @@ func bindViper(cmd *cobra.Command) error {
 	if err := viper.BindEnv("kubeconfig", "KUBECONFIG"); err != nil {
 		return err
 	}
+	if err := viper.BindEnv("client-max-recv-msg-size", "CLIENT_MAX_RECV_MSG_SIZE"); err != nil {
```
wwitzel3, line 195: You don't need the extra binding here. We want the value prefixed with `OCTANT_` you only need to use BindEnv for values that don't have that prefix

nodece: got

wwitzel3:
Looks good to me, someone else PTAL

GuessWhoSamFoo (approved):

GuessWhoSamFoo:
Missing changelog, but we can add one in later.

PR #1239, opened 2020-08-11
Add full label selector to store#Key struct

### LINKED ISSUE #1201: Key#Selector should support searching for objects where a label exists, regardless of label value
Currently, when using `Key` with the `Selector` field, you have to set a value for the key. This is a problem when you want to fetch _all_ objects where that label exists.

This is supported by the K8s API, but in a part of the API that isn't directly used by `Key`.

Currently, `Key` uses `labels.Set` (which is defined as `type Set map[string]string`), and later on, in, e.g., `DynamicCache#listFromInformer` (`internal/objectstore/dynamic_cache.go:281`), we do this:

```
var selector = kLabels.Everything()
if key.Selector != nil {
    selector = key.Selector.AsSelector()
}
```

If it was possible to use [`kLabels.Selector`](https://pkg.go.dev/k8s.io/apimachinery@v0.18.6/pkg/labels?tab=doc#Selector) directly, it would be possible to do something like this:

```
newSelector := kLabels.NewSelector()
requirement, err := kLabels.NewRequirement("app", selection.Exists, []string{})
if err != nil {
	return nil, false, err
}
newSelector.Add(*requirement)
```

I'm not sure if the ideal solution is refactoring `Key` to use `kLabels.Selector` directly in `Key`, or if we should build a `Selector` from `Key#Selector`, and make assumptions that if, e.g., the value is an empty string, that means we're intending to use `selection.Exists` in the query.

N/A

**Environment:**

- Octant version (use `octant version`): `2e4446a6`
- Kubernetes version (use `kubectl version`): `v0.19.0-alpha.3`
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 10.15.6

ipsi:
Support using all types of label matching in selector, such as exists,
contains, not-exists, etc.

- Fixes #1201

**Release note**:
```
* Expose full selector capabilities through `Key` object
```

### DIFF
--- a/internal/objectstore/dynamic_cache.go
+++ b/internal/objectstore/dynamic_cache.go
@@ -8,7 +8,6 @@ package objectstore
 import (
 	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"io"
 	"sync"
@@ -29,6 +28,8 @@ import (
 	kcache "k8s.io/client-go/tools/cache"
 	kretry "k8s.io/client-go/util/retry"
 	sigyaml "sigs.k8s.io/yaml"
+
+	"github.com/pkg/errors"
 
 	"github.com/vmware-tanzu/octant/internal/cluster"
 	"github.com/vmware-tanzu/octant/internal/log"
@@ -278,9 +279,22 @@ func (dc *DynamicCache) listFromInformer(ctx context.Context, key store.Key) (*u
 		l = informer.Lister().ByNamespace(key.Namespace)
 	}
 
+	if key.Selector != nil && key.RequirementsSelector != nil {
+		return nil, false, fmt.Errorf("must provide only one of Key.Selector and Key.RequirementsSelector")
+	}
+
 	var selector = kLabels.Everything()
 	if key.Selector != nil {
 		selector = key.Selector.AsSelector()
+	} else if key.RequirementsSelector != nil {
+		selector = kLabels.NewSelector()
+		for _, r := range *key.RequirementsSelector {
+			requirement, err := kLabels.NewRequirement(r.Key, r.Operator, r.StrValues)
+			if err != nil {
+				return nil, false, errors.Wrap(err, "unable to convert requirement to k8s requirement")
+			}
+			selector = selector.Add(*requirement)
+		}
 	}
 
 	objects, err := l.List(selector)
--- a/internal/objectstore/dynamic_cache.go
+++ b/internal/objectstore/dynamic_cache.go
@@ -8,6 +8,7 @@ package objectstore
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"sync"
@@ -28,8 +29,6 @@ import (
 	kcache "k8s.io/client-go/tools/cache"
 	kretry "k8s.io/client-go/util/retry"
 	sigyaml "sigs.k8s.io/yaml"
-
-	"github.com/pkg/errors"
 
 	"github.com/vmware-tanzu/octant/internal/cluster"
 	"github.com/vmware-tanzu/octant/internal/log"
@@ -279,21 +278,17 @@ func (dc *DynamicCache) listFromInformer(ctx context.Context, key store.Key) (*u
 		l = informer.Lister().ByNamespace(key.Namespace)
 	}
 
-	if key.Selector != nil && key.RequirementsSelector != nil {
-		return nil, false, fmt.Errorf("must provide only one of Key.Selector and Key.RequirementsSelector")
+	if key.Selector != nil && key.LabelSelector != nil {
+		return nil, false, fmt.Errorf("must provide only one of Key.Selector and Key.LabelSelector")
 	}
 
 	var selector = kLabels.Everything()
 	if key.Selector != nil {
 		selector = key.Selector.AsSelector()
-	} else if key.RequirementsSelector != nil {
-		selector = kLabels.NewSelector()
-		for _, r := range *key.RequirementsSelector {
-			requirement, err := kLabels.NewRequirement(r.Key, r.Operator, r.StrValues)
-			if err != nil {
-				return nil, false, errors.Wrap(err, "unable to convert requirement to k8s requirement")
-			}
-			selector = selector.Add(*requirement)
+	} else if key.LabelSelector != nil {
+		selector, err = metav1.LabelSelectorAsSelector(key.LabelSelector)
+		if err != nil {
+			return nil, false, err
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/store/store.go
```diff
@@ -49,11 +51,19 @@ type Store interface {
...
-	Selector   *labels.Set `json:"selector"`
+	Namespace            string         `json:"namespace"`
+	APIVersion           string         `json:"apiVersion"`
+	Kind                 string         `json:"kind"`
+	Name                 string         `json:"name"`
+	Selector             *labels.Set    `json:"selector"`
+	RequirementsSelector *[]Requirement `json:"requirementsSelector"`
+}
+
+// See vendor/k8s.io/apimachinery/pkg/labels/selector.go:120
```
wwitzel3, line 62: Why duplicate this here? We already import labels. Is it just so we can annotate it?

wwitzel3:
. The String method for store.Key should print requirements as well.

pkg/store/store.go
ipsi: All the fields in `labels#Requirement` are (a) private, and (b) not annotated, and it _seemed_ to have issues when I tried it. I'm happy to give it another go, though.

ipsi: Yeah, just checked again, and it silently ignores the values in the JSON - it creates an array with a single object, but all the values in that object are `nil`/`""`/etc. Unsure if that's because they're private or because they're not annotated?

pkg/store/store.go
```diff
@@ -49,11 +51,19 @@ type Store interface {
...
-	APIVersion string      `json:"apiVersion"`
-	Kind       string      `json:"kind"`
-	Name       string      `json:"name"`
-	Selector   *labels.Set `json:"selector"`
+	Namespace            string         `json:"namespace"`
+	APIVersion           string         `json:"apiVersion"`
+	Kind                 string         `json:"kind"`
+	Name                 string         `json:"name"`
+	Selector             *labels.Set    `json:"selector"`
+	RequirementsSelector *[]Requirement `json:"requirementsSelector"`
```
scothis, line 59: slices are nil-able

```suggestion
	RequirementsSelector []Requirement `json:"requirementsSelector"`
```

internal/objectstore/dynamic_cache.go
```diff
@@ -278,9 +279,22 @@ func (dc *DynamicCache) listFromInformer(ctx context.Context, key store.Key) (*u
...
+
 	var selector = kLabels.Everything()
 	if key.Selector != nil {
 		selector = key.Selector.AsSelector()
+	} else if key.RequirementsSelector != nil {
+		selector = kLabels.NewSelector()
+		for _, r := range *key.RequirementsSelector {
+			requirement, err := kLabels.NewRequirement(r.Key, r.Operator, r.StrValues)
+			if err != nil {
+				return nil, false, errors.Wrap(err, "unable to convert requirement to k8s requirement")
```
scothis, line 294: no need to use a library

```suggestion
				return nil, false, fmt.Errorf("unable to convert requirement to k8s requirement: %w", err)
```

pkg/store/store.go
scothis, line 62: the line number will go stale over time. If the reference is useful, link to a specific commit, otherwise remove the reference.

scothis: > All the fields in labels#Requirement are (a) private, and (b) not annotated, and it seemed to have issues when I tried it.

That's usually a sign that you're trying to use sometime in a way it wasn't intended to be used.

Have you looked at https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#LabelSelector

ipsi: That looks like it would work, but it's not a particularly nice API, as there are two different ways of searching for `key=value`, one of which duplicates what we already have in `Key#Selector`.

`equals`:
```
{
  "matchLabels": { "label1":"test" }
}
```
OR
```
{
  "matchExpressions": [{"key":"label1", "operator":"in", "values":["test"]}]
}
```

`exists`:
```
{
  "matchExpressions": [{"key":"label1", "operator":"exists", "values":[]}]
}
```

ipsi: It also doesn't support the full set of operators, missing greater-than and less-than operators, though I'm not sure how much use they'd see in practice?

ipsi:
I think I also need to modify the `Key#Validate` method, regardless of exactly which struct we use for the data, as it's not valid for, e.g., an `Exists` operator to have values, or for an `In` operator to **not** have values.

pkg/store/store.go
ipsi: I've implemented it with that struct, and I can push that out if you're interested in seeing it. Quirks:

* Have to write our own `String()` function for it
* Have to convert *both* `matchLabels` and `matchExpressions` to a String, which produces an awkward result like `CacheKey[Namespace='default', APIVersion='v1', Kind='ConfigMap', LabelSelector='MatchLabels=foo1=bar1,foo4=bar4, MatchExpressions=foo2 In (bar2),foo3 Exists']`

Conversion is at least handled by `metav1.LabelSelectorAsSelector`, so apart from the API design and the `String()` function, it's pretty simple.

ipsi:
Looks like the selector is used in two places - `DynamicCache#listFromInformer` and `DynamicCache#listFromDynamicClient` - missed the second one.

pkg/store/store.go
scothis: The core types are not always the best example for good greenfield design , but they are well understood in the community. https://godoc.org/k8s.io/api/apps/v1#DeploymentSpec

> It also doesn't support the full set of operators, missing greater-than and less-than operators

Greater than and less than operators on strings are always a bit dodgy unless you also define the character set, locale,  lexicographical rules, etc.

ipsi: That's a fair point. I've pushed a commit using `LabelSelector` instead of the custom type. Might still be some tweaks needed, especially to the `String` function.

internal/objectstore/dynamic_cache.go
```diff
@@ -279,21 +278,18 @@ func (dc *DynamicCache) listFromInformer(ctx context.Context, key store.Key) (*u
...
-	} else if key.RequirementsSelector != nil {
-		selector = kLabels.NewSelector()
-		for _, r := range *key.RequirementsSelector {
-			requirement, err := kLabels.NewRequirement(r.Key, r.Operator, r.StrValues)
-			if err != nil {
-				return nil, false, errors.Wrap(err, "unable to convert requirement to k8s requirement")
-			}
-			selector = selector.Add(*requirement)
+	} else if key.LabelSelector != nil {
+		fmt.Println("Key: " + key.String())
```
scothis, line 289: debugging code?

internal/objectstore/dynamic_cache.go
```diff
@@ -317,6 +313,13 @@ func (dc *DynamicCache) listFromDynamicClient(ctx context.Context, key store.Key
 	var selector = kLabels.Everything()
 	if key.Selector != nil {
 		selector = key.Selector.AsSelector()
+	} else if key.LabelSelector != nil {
+		fmt.Println("Key: " + key.String())
```
scothis, line 317: debugging code?

PR #1240, opened 2020-08-11
Add no-op handler for loading action

GuessWhoSamFoo:
The uploader component must check if the backend is in loading state each time the component is created. This PR creates a no-op handler to prevent (`received websocket unknown message of type event.octant.dev/unknown with {message: "unknown request action.octant.dev/loading", payload: {…}}`) from spamming the console.

### DIFF
--- a/internal/api/content_manager.go
+++ b/internal/api/content_manager.go
@@ -188,6 +188,10 @@ func (cm *ContentManager) Handlers() []octant.ClientRequestHandler {
 			RequestType: action.RequestSetNamespace,
 			Handler:     cm.SetNamespace,
 		},
+		{
+			RequestType: CheckLoading,
+			Handler:     cm.Loaded,
+		},
 	}
 }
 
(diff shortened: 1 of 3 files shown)


GuessWhoSamFoo:
Added fix to tests breaking

pkg/plugin/api/api.go
```diff
@@ -55,8 +55,13 @@ func (a *grpcAPI) Start(ctx context.Context) error {
 		service: a.Service,
 	}
 
+	size := viper.GetInt("client-max-recv-msg-size")
+	if size == 0 {
```
wwitzel3, line 59: If we want to prevent zero as a size, then we should do it in the flag setup. I believe we can set min max thresholds

GuessWhoSamFoo: If `api.Start` is called from a test environment where flags are not setup, setting a default earlier would not be read unless the size is passed down rather than fetched through `GetInt`. I ended up setting a viper default in the test although we will need to remember to pass the size explicitly instead if there is ever a use case for creating `NewDashboardClient` externally.

PR #1242, opened 2020-08-11
Dynamic components

bryanl:

This change is an attempt to start moving from the switch case plus component rendering system to something more dynamic. We'll need this hook to support the ability for plugins to have their own components. While deciding on a path, I noticed that rendering components felt slower than necessary, so I started at the beginning to understand what was going on. 

The first thing I noticed that was our usage of the discovery client was taxing the cluster. This manifested in a throttle message that was issued by client-go periodically.

From there I re-evaluated how generating components works. The good news is that in the overwhelming majority of cases, Octant generates a component response in less than 30ms on my workstation. I noticed that in certain cases we were not stopping the polling of a path when the websocket client went away. To fix this, I added a new channel that we can close explicitly to stop the poller.

Our slowness is all in the front end. From my benchmarking, the overview page can take 3500+ ms to render. With further research, I noticed that Clarity datagrids can take 150-200ms to render. In a cursory pass, I could not find a way to make this faster, but since this was not the goal, I moved on.

Previously, Octant used a case statement in multiple locations to render components in a dynamic fashion. Since components are not known ahead of time, this is how we compensated. This change removes that concept and introduces a new method based on Angular's component factory mechanism. Instead of iterating through a switch, instead, Octant now creates an instance of a component and inserts it into the page dynamically. This doesn't appear to be any slower than what we had before, but it does give us more flexibility around what we can add.

All view components now inherit from a common base, `AbstractViewComponent`. This component handles the basics of view management. This means that view components can only focus on their business logic.

This change includes:

- Use cached discovery client. This should solve the `Throttled` messages printed to standard err
- Ensure websocket client stops when reader connection ends. This should make Octant not have dangling websockets connections that send to no where.
- Move to dynamic components. Components are now loaded dynamically instead of using the content switcher and its long switch case.


mklanjsek:
This is dope 🎆

mklanjsek (changes requested):
This is awesome but bunch of things are broken! I noted a few small comments that require code changes.

The biggest problem is that I see views being updated every second and on top of performance degradation, in some cases that's causing flashing (see testing notes below). It would be good if the update logic can be standardized and puled out to new base component class, to make sure we update only stuff that changed.

Testing notes:
- The infamous `Unable to preventDefault inside passive event listener invocation.`  error is back in console inside the ResourceViewer and YAML view,
- Terminal flashes and it's pretty much unusable. Additionally, opening terminal produces this error in console:
```main-es2015.a542a1464d307b58d51b.js:1 ERROR Error: This API only accepts integers
    at e._verifyIntegers (1-es2015.2486becf66a916233d9d.js:1)
    at e.resize (1-es2015.2486becf66a916233d9d.js:1)
    at e.fit (1-es2015.2486becf66a916233d9d.js:1)
    at 1-es2015.2486becf66a916233d9d.js:1...
```
- Opening Logs tab on the other hand makes following errors show up in console (on top of constant ):
```ERROR TypeError: Cannot read property 'subscribe' of undefined
    at e.loadView (1-es2015.2486becf66a916233d9d.js:1)
    at e.ngOnChanges (1-es2015.2486becf66a916233d9d.js:1)...
```
- Yaml editor seems to be in read-only mode, same in Apply Yaml panel.

- In general, it would be good if we all have same linter settings - there is a lot of formatting changes in html and scss files that make review harder and produce unnecessary file versions.

build.go
```diff
@@ -277,7 +277,7 @@ func build() {
 	if runtime.GOOS == "windows" {
 		artifact = "octant.exe"
 	}
-	runCmd("go", nil, "build", "-o", "build/"+artifact, GO_FLAGS, "-v", "./cmd/octant")
+	runCmd("go", nil, "build", "-mod=vendor", "-o", "build/"+artifact, GO_FLAGS, "-v", "./cmd/octant")
```
scothis, line 280: shouldn't be necessary with go 1.14+

bryanl: I thought the same, but was wondering why Go was no pulling the proper set of dependencies.

scothis: the go.mod file was set to 1.13, and is now 1.15 on master

bryanl:
@mklanjsek I believe all the issues have been addressed. The loading indicator should be revisited in a future change, but it should not be broken now.

wwitzel3:
This works well for me. I tested terminals, logging, port forwards, and resource viewer.
The load time is WAY faster on the initial load and after the initial load, even on a fairly large overview, coming back to it from a summary page is almost instant.

wwitzel3 (approved):

mklanjsek:
Looks good to me as well, nicely done! Just a couple of issues that I run into:
- Datagrid initial update is much better now (although I feel we should put a spinner inside the datagrid until it receives the data), but all updates after that are confusing. We should not make data disappear from grids on updates.
  
- I am seeing a bunch of `No component factory` errors in Storybook. They get dumped to console when switching between components, but the easiest way to repro is to select the List component: nothing shows up in Canvas and following error is in console:
```
Error: No component factory found for DatagridComponent. Did you add it to @NgModule.entryComponents?
    at noComponentFactoryError (core.js:19825)```

mklanjsek (approved):

PR #1247, opened 2020-08-12
Decouple ThemeService from Monaco

scothis:

The ThemeService is now fully decoupled from Monaco. Consumers of Monaco
do not need to take any special steps for the editor to get the correct
theme. More importantly, loading the ThemeService will no longer also
load Monaco.

The ThemeService publishes change events for any interested consumer
when the loaded theme changes.

I can't find a hook in ng-monaco-editor that will allow us to define/update
the desired theme for the editor be Monaco is initialized, so we still eagerly
initialize the editor in the InitService.

**Release note**:
```
none
```

### DIFF
--- a/web/src/app/components/smart/home/home.component.spec.ts
+++ b/web/src/app/components/smart/home/home.component.spec.ts
@@ -1,36 +1,28 @@
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 
 import { HomeComponent } from './home.component';
-import {
-  MonacoEditorComponent,
-  MonacoProviderService,
-  MonacoEditorConfig,
-} from 'ng-monaco-editor';
-import { themeServiceStub } from 'src/app/testing/theme-service-stub';
-import { ThemeService } from 'src/app/modules/sugarloaf/components/smart/theme-switch/theme-switch.service';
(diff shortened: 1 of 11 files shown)


mklanjsek:
Another thing I noticed and it might be related to above issue - when you manually delete the theme from local storage, Octant updates and toggles theme.

scothis:
> I was testing running Octant in two windows, when suddenly in the middle of the update one of them started flashing like this

I haven't seen that flicker. I have seen the theme switch be a little slow to apply when a window is also rendering a new page. Complex pages like the Namespace Overview can take several hundred milliseconds to several seconds to render blocking other work. (We should dig into the perf, because that's quite slow). But that's not what you're seeing.

> when you manually delete the theme from local storage, Octant updates and toggles theme

This is expected and a bit surprising. #1214 added a localStorage mutation listener to sync the theme state between windows. So if you're in dark mode, removing the key I would expect to going back to the default (light) theme. What's surprising, at least initially, is that it doesn't switch to the default theme, but toggles the theme. This is a side effect of how I do the check when receiving the storage event. If the new value is not the current theme, then the theme is toggled. I like the naive safety of the current implementation because an invalid theme will be updated to a valid theme without any special validation. If we ever have more than two themes, this behavior will need to change.

PR #1252, opened 2020-08-14
Display a 404-like page if a resource is not found

### LINKED ISSUE #422: Gracefully handle situation when crd is not installed
If a CRD is not installed, Octant is upset. Instead of erroring, create a method to inform the user of the issue and a possible solution.

```
err in navigation: generate entries for Custom Resources: listing custom resources for "httpproxies.projectcontour.io": client resource: no matches for kind "HTTPProxy" in group "projectcontour.io"2019-11-13T09:14:37.207-0500       ERROR   api/navigation_manager.go:97    load namespaces {"poller-name": "navigation", "err": "generate entries for Custom Resources: listing custom resources for \"httpproxies.projectcontour.io\": client resource: no matches for kind \"HTTPProxy\" in group \"projectcontour.io\""}
github.com/vmware-tanzu/octant/internal/api.(*NavigationManager).runUpdate.func1
        /Users/bryan/Development/vmware/octant/internal/api/navigation_manager.go:97
github.com/vmware-tanzu/octant/internal/api.(*job).run.func1
        /Users/bryan/Development/vmware/octant/internal/api/poller.go:111
```

scothis:

Previously, Octant would redirect to a parent path segment when a
resource is not found. This could be because the path is not mapped in
Octant, the particular namespace or resource is not found, or in the
case of CRDs, the CRD is not defined. This behavior means that the URL a
user navigated to may not be the URL that is rendered.

Now, the URL is preserved and an error page is displayed instead. The
user is present with a small number of options including going to the
parent resource, or waiting. If the user remains on the page and the
resource becomes valid, the new resource will replace the error page.
This behavior is particularly useful when watching a single resource
that comes and goes.

- Fixes #422

The formatting of the error page could use another iteration, but I wanted to open this to solicit feedback on the concept.

The below was captured when viewing, deleting and recreating a resource.

Screen Recording 2020-08-14 at 3 31 47 PM

**Release note**:

```
Render a 404-style error page when a resource is not found
```


wwitzel3:
It is too jarring in dark mode to get that white background of Wikipedia injected in to the page. I think we might need to look at other options for fun not found entries.

image

scothis:
@wwitzel3 updated. We can make it pretty over time

Screen Shot 2020-08-20 at 4 52 37 PM
Screen Shot 2020-08-20 at 4 52 34 PM

PR #1299, opened 2020-08-27
Add a ButtonGroup to the Table

scothis:

Tables support grid actions that perform actions on a given row of the
table, but did not have a way to take an action on the whole table. Now
an optional button group is rendered to the right of the title.

For example, a button could be used to create a new item in the table.

Screen Shot 2020-08-27 at 1 13 28 PM

**Release note**:
```
Add an optional button group to data grid tables
```

### DIFF
--- a/pkg/view/component/table.go
+++ b/pkg/view/component/table.go
@@ -12,6 +12,9 @@ import (
 	"sync"
 
 	"github.com/davecgh/go-spew/spew"
+	"github.com/pkg/errors"
+
+	"github.com/vmware-tanzu/octant/pkg/action"
 )
 
 // TableFilter describer a text filter for a table.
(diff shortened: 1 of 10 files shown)


pkg/view/component/table.go
```diff
@@ -27,6 +28,43 @@ type TableConfig struct {
 	EmptyContent string                 `json:"emptyContent"`
 	Loading      bool                   `json:"loading"`
 	Filters      map[string]TableFilter `json:"filters"`
+	ButtonGroup  *ButtonGroup           `json:"buttonGroup,omitempty"`
```
GuessWhoSamFoo, line 31: For consistency, can we also have a method like:

```
func (t *Table) AddButtonGroup(buttonGroup *ButtonGroup) {
	t.Config.ButtonGroup = buttonGroup
}
```

web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.html
```diff
@@ -1,4 +1,11 @@
-<h4 *ngIf="showTitle()">{{ title }}</h4>
+<div class="clr-row clr-align-items-center clr-justify-content-between" *ngIf="showTitle() || buttonGroup">
```
bryanl, line 1: Why did you choose this method over Clarity's provided clr-dg-action-bar?

wwitzel3: Example here https://clarity.design/documentation/datagrid/batch-action

scothis: The short answer is that I wasn't aware of `clr-dg-action-bar`. The component itself [doesn't do much](https://github.com/vmware/clarity/blob/c77dee037dda21556dba30aa4eb742ea1832677b/packages/angular/projects/clr-angular/src/data/datagrid/datagrid-action-bar.ts), it's just a semantic concept.

Added, but keeping it outside of `clr-datagrid` to preserve the basic layout of the button to the side of the title. By default the buttons appear below the title, which is odd to my eyes, but that opinion is weakly held.

pkg/view/component/table.go
scothis: Added an `AddButton` method instead. It will append to an existing button group or create a new one if nil.

PR #1302, opened 2020-08-28
Fix edit service to have sorted selectors

GuessWhoSamFoo:
Navigating into Service then editing the configuration shows a list of selectors that are not sorted from the backend and can appear in different order each content response.

Also moves the datagrid placeholder and column drawing outside of `setTimeout` to prevent flickering on component redraw.

### DIFF
--- a/internal/printer/service.go
+++ b/internal/printer/service.go
@@ -8,6 +8,7 @@ package printer
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strings"
 
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


scothis:
A bit off topic, but should the `setTimeout` call instead be `requestAnimationFrame`?

mklanjsek:
Didn't measure it, but I see the slowdown as well.
+1 for adding spinners inside the datagrid.
It would be even better if we can know the exact number of elements inside datagrids on initial update to prevent grid resize, but I don't think that's possible.

GuessWhoSamFoo:
This PR is now updated to only prevent the unsorted list of selectors from moving when editing a service. Flickering is handled in #1315

wwitzel3 (approved):

bryanl (approved):

PR #1305, opened 2020-08-30
Add modal component

GuessWhoSamFoo:
This PR adds a [modal component](https://clarity.design/documentation/modals). It is intended to handle three cases:

1. Display some content in a modal
2. Display a form
3. Display content and a form

If the modal does not contain a form, the only button available is `Close` (the modal is always closable in all cases). If a form is added to the modal component, the buttons are `Submit` and `Cancel` where submit validates the form before sending a message with action.

A modal can be opened via a button or programatically though modifying the modal config as open.

Example with a stepper in a modal to be opened via table with buttonGroup:
image

Code sample:
```
fft := component.NewFormFieldText("textFormLabel", "textFormName", "")
fft.AddValidator("placeholder", "this is an error", []string{"required"})

form := component.Form{
    Fields: []component.FormField{
        fft,
    },
}

stepper := component.NewStepper("stepper", "test")
stepper.AddStep("Step1", form, "title1", "description1")
stepper.AddStep("Step2", form, "title2", "description2")

modal := component.NewModal(component.TitleFromString("Modal example"))
modal.SetBody(stepper)
modal.SetSize(component.ModalSizeLarge)

button := component.NewButton("Click me", action.Payload{}, component.WithModal(modal))


table := component.NewTable("test", "test", []component.TableCol{})
table.Config.ButtonGroup = component.NewButtonGroup()
table.Config.ButtonGroup.AddButton(button)
```

### DIFF
--- a/web/src/app/modules/shared/components/presentation/modal/modal.component.html
+++ b/web/src/app/modules/shared/components/presentation/modal/modal.component.html
@@ -4,119 +4,24 @@
   </h3>
   <div class="modal-body">
     <app-view-container [view]="body"></app-view-container>
-    <form clrForm *ngIf="form" (ngSubmit)="onFormSubmit()" [formGroup]="formGroup">
-      <div *ngFor="let field of form.fields; trackBy: trackByFn">
-        <ng-container [ngSwitch]="field.type">
-          <ng-container *ngSwitchCase="'checkbox'">
-            <clr-checkbox-container>
-              <label [for]="field.name">{{ field.label }}</label>
-              <clr-checkbox-wrapper
-                *ngFor="let opt of fieldChoices(field); trackBy: trackByFn"
-              >
-                <input
-                  type="checkbox"
-                  clrCheckbox
-                  [formControlName]="field.name"
-                  [value]="opt.value"
-                  [checked]="opt.checked"
-                />
-                <label>{{ opt.label }}</label>
-              </clr-checkbox-wrapper>
-              <clr-control-error>{{ field.error }}</clr-control-error>
-            </clr-checkbox-container>
-          </ng-container>
-          <ng-container *ngSwitchCase="'radio'">
-            <clr-radio-container>
-              <label [for]="field.name">{{ field.label }}</label>
-              <clr-radio-wrapper
-                *ngFor="let opt of fieldChoices(field); trackBy: trackByFn"
-              >
-                <input
-                  type="radio"
-                  clrRadio
-                  [formControlName]="field.name"
-                  [value]="opt.value"
-                />
-                <label>{{ opt.label }}</label>
-              </clr-radio-wrapper>
-              <clr-control-error>{{ field.error }}</clr-control-error>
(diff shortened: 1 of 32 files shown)


mklanjsek:
This is very cool! My only comment is that it would be good to add more storybook stories to illustrate modal opening and closing as well as interactivity inside the modal.

GuessWhoSamFoo:
Rebased to be using angular 10 and added a form example/

web/src/app/modules/shared/components/presentation/modal/modal.component.html
```diff
@@ -0,0 +1,122 @@
+<clr-modal [(clrModalOpen)]="opened" [clrModalStaticBackdrop]="false" [clrModalClosable]="true" [clrModalSize]="size">
+  <h3 class="modal-title">
+    <app-view-title [views]="title"></app-view-title>
+  </h3>
+  <div class="modal-body">
+    <app-view-container [view]="body"></app-view-container>
+    <form clrForm *ngIf="form" (ngSubmit)="onFormSubmit()" [formGroup]="formGroup">
+      <div *ngFor="let field of form.fields; trackBy: trackByFn">
+        <ng-container [ngSwitch]="field.type">
```
bryanl, line 9: The code / html that renders forms should be in a separate component.

web/src/app/modules/shared/components/presentation/modal/modal.component.html
```diff
@@ -0,0 +1,122 @@
...
+          <ng-container *ngSwitchDefault>
+            Unable to display form field type {{ field.type }}
+          </ng-container>
+        </ng-container>
+        <br>
+      </div>
+    </form>
+  </div>
+  <div class="modal-footer">
+    <ng-container *ngIf="form else closeButton">
```
bryanl, line 111: Would the modal component be more useful if it gave more choice to the user? For example, let the user decide which buttons they want to show and what should happen if they show those buttons. For code that creates a modal component, create helpers that make common tasks trivial.

bryanl (changes requested):

web/src/app/modules/shared/components/presentation/modal/modal.component.html
GuessWhoSamFoo: Form is now extracted from card and summary components. This cannot be applied to the stepper because there is validation between each step and it is simpler to let the clarity code do the lifting.

web/src/app/modules/shared/components/presentation/modal/modal.component.html
GuessWhoSamFoo: The modal now supports three cases of usage. Forms still have a close or submit buttons. For all other content, the user can either choose to provide button(s) or omit adding any buttons entirely where a close button will fill instead.

wwitzel3 (approved):
I think this is fine for now. I'm still a little not sure about the magical button behavior, but I think we can iterate on that and it isn't a blocker for this PR.

PR #1313, opened 2020-09-01
Fixing safari height bug (#1268)

### LINKED ISSUE #1268: Header should grow to fit content in narrow windows

Open Octant in a narrow, but not tiny window. The header content overflows into the body area. This looks strange in dark mode and the content is nearly invisible in light mode.

All of the header content should be in the header.

Screen Shot 2020-08-20 at 4 52 37 PM
Screen Shot 2020-08-20 at 4 52 34 PM

**Environment:**

- Octant version (use `octant version`): 0.15 and HEAD
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Safari (macOS 10.15)

alexbarbato:
Safari doesn't expand header as screen shrinks.

- Fixes #1268

Not in love with this fix as its kind of hack due to the dependencies on clarity. 

There is an inherit dependency that the height of flex items within the container don't change. Granted, with this fix at least the change would look the same on all browsers.

**Release note**:
```
Fixing safari height bug in the header so all browsers render the header the same.
```

### DIFF
--- a/changelogs/unreleased/1313-alexbarbato
+++ b/changelogs/unreleased/1313-alexbarbato
@@ -1,0 +1,1 @@
+Fixing safari height bug in the header so all browsers render the header the same.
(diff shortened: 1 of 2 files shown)


alexbarbato:
Whoops - sorry about that.

Amended 👍

PR #1315, opened 2020-09-02
reduce flicker when components update

bryanl:

This change reduces flicker when components update by reusing existing resources and reducing the number of times static items change.

- In view-container:
  - Replace on changes hook with a setter that directly invokes loadView. 
  - In loadView use existing component ref if it exists and creates one if it does not.
- In list.component:
  - Update the change detection to look at the items directly. No changes could be updated since the ref to the array items is in did not change. 
  - Use an items property instead of referring to the view itself.
- In datagrid.component:
  - disable the loading indicator for datagrids until the loading timeout rxjs can be solved.


mklanjsek:
This looks good to me - I don't see flicker and performance is pretty good.

Would be good to get more eyes on it.

wwitzel3:
This change is working well for me.

wwitzel3 (approved):

GuessWhoSamFoo (approved):
Looks fine after manual review of components generated via factory (found issues, but unrelated)

PR #1341, opened 2020-09-13
Create a bottom panel to show secondary content

bryanl:

This set of changes creates a bottom panel component to show secondary content. This change is meant to supersede the functionality provided in the SliderView component by introducing the following features:

- Only includes a bottom panel and not the contents of the bottom panel. This reduces coupling and allows the parent component to decide what is shown. 
- Allows for resizing the content without using the "ghost" method that was previously employed
- Is designed to live in the same container as other content rather than as an overlay. ie. It will push up content when it is activated. 

This change also includes:
- Temporarily makes the collapsed navigation view scrollable. 

The panel is included, but it is not activated. To see it working, set `isBottomPanelEnabled` to true in ContainerComponent and add some content to `app-bottom-panel` in the view.

### DIFF
--- a/web/src/app/modules/shared/components/smart/bottom-panel/bottom-panel.component.html
+++ b/web/src/app/modules/shared/components/smart/bottom-panel/bottom-panel.component.html
@@ -1,0 +1,20 @@
+<div
+  class="slider-container"
+  mwlResizable
+  [resizeCursors]="resizeCursors()"
+  [allowNegativeResizes]="true"
+  (resizing)="updateSliderPosition($event)"
+>
+  <div
+    class="gutter"
+    mwlResizeHandle
+    [resizeEdges]="resizeEdges"
+    [ngClass]="gutterClass()"
+  ></div>
+
+  <a class="slider-container--trigger" (click)="toggle()">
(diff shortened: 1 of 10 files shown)


mklanjsek:
I like this, except one thing: IMO collapsed navigation should never be scrollable. Can the slider be positioned inside `content-area` so the navigation still consumes the whole height of the window?

mklanjsek (changes requested):
Love this, but I am seeing following problems when testing it:
- after navigating from one page to another, the bottom panel collapses to its minimized state - I would expect it to stay open, 
- the compressed side nav is broken after scrollbar has been added - it's impossible to select any items in flyout menu. The flyout menu shows up briefly, but then disappears after pointer is moved over the scrollbar,
- As commented above, I still think that bottom panel has to be be positioned inside the main pane and left navigation should go all the way to the bottom of the window,

bryanl:
@mklanjsek 

* Yes, the panel will collapse. It's not fully integrated as of this commit, so it is disabled. I have code that'll fix that in a future commit.
* I believe this bug is in the compressed size nav code. The menu isn't updating itself properly, and this will need to be fixed before we enable this new feature.

mklanjsek (approved):

PR #1360, opened 2020-09-18
use icon and color for indicator to make it more accessible

### LINKED ISSUE #1335: Please make service up/down indicators colorblind friendly
For colorblind folks, it's hard to tell the difference between red/green here:
image

Can we do an icon instead of color?

wwitzel3:

image

image

- Fixes #1335


web/src/app/modules/shared/components/presentation/indicator/indicator.component.html
```diff
@@ -1,7 +1,11 @@
 <ng-container *ngIf="currentStatus">
   <ng-container *ngIf="detail !== undefined; else showStatus">
     <clr-signpost>
-      <div [ngClass]="[currentStatus, 'indicator']" clrSignpostTrigger></div>
+
```
GuessWhoSamFoo, line 4: Since the indicator class is removed, the scss can be removed entirely.

GuessWhoSamFoo (commented):
For an error status, I wonder if `times-circle` is easier to understand (especially if the color is ignored)

web/src/app/modules/shared/components/presentation/indicator/indicator.component.html
```diff
@@ -1,7 +1,11 @@
...
+      <clr-icon *ngIf="currentStatus === 'error'" shape="exclamation-circle" class="is-error is-solid" size="16" clrSignpostTrigger></clr-icon>
+      <clr-icon *ngIf="currentStatus === 'warning'" shape="info-circle" class="is-warning is-solid" size="16" clrSignpostTrigger></clr-icon>
+      <clr-icon *ngIf="currentStatus === 'success'" shape="check-circle" class="is-success is-solid" size="16" clrSignpostTrigger></clr-icon>
+
```
GuessWhoSamFoo, line 8: Update ng-template below as well

wwitzel3:
I tried times-circle but it didn't look right (scaling) and goes against what a lot of other systems do.

mklanjsek:
Also, please add extra 2-4px spacing between icon and text to improve readability.

It's looking much better now, but we should consider replacing the signpost with tooltip at some point.

wwitzel3:
SCSS remove and padding added.

wwitzel3:
image

GuessWhoSamFoo:
image

image

It's also interesting that our `is-warning` color is different from the documentation. Did we miss something from https://github.com/vmware/clarity/pull/3790 ?

mklanjsek:
These are very nice @GuessWhoSamFoo,  but I would still prefer the current icons for consistency. 

Also more importantly, if we use same icon for all 3 states, that does not solve the accessibility issue.

GuessWhoSamFoo (approved):

jongio:
This is AWESOME.  Thank you so much for doing this.  Certaintly going to make my Octant experience better

PR #1367, opened 2020-09-20
Move apply YAML to modal; added shortcut key

GuessWhoSamFoo:
This PR changes the apply YAML feature to be used in a modal instead of a routerLink. It will allow users to use this feature without context switching or waiting for a new page to load.

image

Notes: 
 - The URI of a monaco editor is now generated with a random string to allow multiple editors in the DOM at once. Previously there would be a collision as the URI was hardcoded with the error: `Cannot add model because it already exists`
 - Changed key binding event listener to keydown in order to prevent potential browser shortcut conflicts.
 - It is possible to have modals stacked on top of each other (e.g. `Ctrl + /` then `Ctrl + y`). This can be fixed after #1305 through a modal service.


web/src/app/modules/shared/components/smart/helper/helper.component.html
```diff
@@ -92,6 +92,16 @@ <h3 class="modal-title">Keyboard Shortcuts</h3>
...
           </clr-stack-content>
         </clr-stack-block>
+        <clr-stack-block
+        [clrStackViewLevel]="2"
+        [clrStackViewSetsize]="0"
+        [clrStackViewPosinset]="1"
+      >
+        <clr-stack-label>Apply YAML</clr-stack-label>
+        <clr-stack-content>
+          <span class="label">Ctrl<span class="badge">Y</span></span>
```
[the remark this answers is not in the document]
GuessWhoSamFoo: I'm inclined to leave it as upper case for aesthetic purposes and show `Shift` explicitly if used.

wwitzel3 (approved):
LGTM

web/src/app/modules/shared/components/smart/editor/editor.component.ts
```diff
@@ -58,6 +59,8 @@ export class EditorComponent extends AbstractViewComponent<EditorView> {
       this.pristineValue = view.config.value;
       this.options.readOnly = view.config.readOnly;
       this.options.language = view.config.language;
+      this.uri =
+        'file:text-' + Math.random().toString(36).substring(2, 15) + '.yaml';
```
scothis, line 63: should this be in the constructor instead? Creating unnecessary churn in the object model can only hurt performance.

GuessWhoSamFoo: Fixing this as part of #1396

Issue #1353, opened 2020-09-17
Left Navigation redesign


mklanjsek:
Growth of Octant plugin ecosystem and recent UI features like #1341 are urging the need for redesign of the Left Navigation. The main problem we are trying to solve is:
- Vertical Navigation doesn’t scroll when collapsed. We need to be able to show more items in that list or divide the items into smaller groups.  
 
**Describe the solution you'd like**
Left Navigation can accommodate large amount of items without using too much screen real estate. Navigation should still support both collapsed and expanded state to satisfy the needs of both novice and advanced users.

**Anything else you would like to add:**
This is intended to be a main issue to track the redesign. Please add any additional problems of current design, feedback and ideas how to improve it in comments below.

mklanjsek:
Here is the last in series of prototypes I created to address this issue:

![out](https://user-images.githubusercontent.com/34245594/94072009-1fef3d80-fdb2-11ea-8320-7a57099b7b56.gif)

Based on discussion from today's community meeting (link), the next iteration will be mix of two above modes:
- there will be tab control on the left with all groups (modules) like in collapsed view in above gif,
- the list of items for each group will be by default expanded, like in expanded view above,
- we should try to eliminate or minimize the collapsed/expand slider,
- the left column with groups should have overflow flyout in case we can not fit all groups inside the current view,

mklanjsek:
Here is the latest prototype based on feedback from yesterday:

![tabgroups-left-expanded](https://user-images.githubusercontent.com/34245594/94195378-27751c00-fe70-11ea-9d06-8600fa85bd22.gif)

For everyone who would like to play with it, code is [here](https://github.com/mklanjsek/octant/pull/new/nav-redesign) (note: it's still on prototype level)

wwitzel3:
I love that.

mklanjsek:
Thanks @wwitzel3! 

Module structure is same as before. Besides, I do not see that level of module info inside the data coming down to client.  How can client figure that out?

mklanjsek:
Here is a quick recording of the latest prototype, @vmware-tanzu/octant-reviewers please let me know what you think before I fix a few remaining issues and submit the PR. 

![navigation_modules](https://user-images.githubusercontent.com/34245594/95213232-d1777100-07ab-11eb-94a6-73a8ada3e526.gif)

[...]

Issue #1359, opened 2020-09-18
Too much scrollbar


wwitzel3:
Current master has too much Scrollbar. Needs less.

![image](https://user-images.githubusercontent.com/27317/93590015-c800a300-f97b-11ea-8142-d83545db56e6.png)

GuessWhoSamFoo:
Closed via #1368

PR #1368, opened 2020-09-21
Restore the collapsed left navigation functionality

mklanjsek:
Changes introduced in #1341 inadvertently broke the left navigation in collapsed state. Namely, the flyout menu was positioned incorrectly and in some cases was not accessible after the scrollbar showed up. This change brings the old functionality back. 

The Navigation component simply does not support scrolling and that will be addressed during the navigation redesign (see #1353), but for the short term (0.16 release) we should keep that functionality the way it was.

### DIFF
--- a/web/src/app/modules/sugarloaf/components/smart/container/container.component.scss
+++ b/web/src/app/modules/sugarloaf/components/smart/container/container.component.scss
@@ -25,9 +25,6 @@
 }
 
 .main-container {
-  flex: 1 1 auto;
-  overflow-y: scroll;
-
   // Navigation's color variables for Light Theme.
 
   :host-context(body) {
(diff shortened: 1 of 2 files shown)


GuessWhoSamFoo (commented):
The main-container and content-container will also need to be reverted as well

mklanjsek:
@GuessWhoSamFoo Left these in since I did not find any issues with them. Please let me know if I missed something.

GuessWhoSamFoo:
See the screeenshot in #1359 

I'm not sure of the browser/OS specifics

mklanjsek:
@GuessWhoSamFoo PTAL

GuessWhoSamFoo:
This breaks the bottom panel temporarily. We can figure out the resizing in 0.17

GuessWhoSamFoo (approved):

PR #1381, opened 2020-09-21
Update quick switcher UI; add namespace to search

GuessWhoSamFoo:
image

This PR improves the quick switcher by:
 - Allowing the user to see current search input while scrolling through results
 - Using case insensitive search (e.g. `Deployment` would yield no results)
 - Moving resource groupings to the right so results are easier to read
 - Adding experimental support for namespace search via starting with `!`

### DIFF
--- a/changelogs/unreleased/1381-GuessWhoSamFoo
+++ b/changelogs/unreleased/1381-GuessWhoSamFoo
@@ -1,0 +1,1 @@
+Updated quick switcher UI and added namespace to search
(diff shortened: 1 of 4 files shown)


wwitzel3:
Let's add something like `!namespace` to the shadow text .. or somewhere. I want that to be discoverable because it is very useful.

GuessWhoSamFoo:
image

Adjusted CSS a bit more so the pointer doesn't flicker when moving between search results.

wwitzel3 (approved):

PR #1403, opened 2020-09-24
Fix since creation option in log viewer

wwitzel3:

Fixes a bug in master where the `Creation` option in the Log selector was not working.

This was due to a JavaScript number rounding issue since this number exceeded Number.MAX_SAFE_INTEGER
image

This was solved by setting the since time to `-1` and using a negative number to trigger looking up the pod creation time and using that to initialize the log streamer.

### DIFF
--- a/internal/api/container_logs.go
+++ b/internal/api/container_logs.go
@@ -32,6 +32,7 @@ type logEntry struct {
 const (
 	RequestPodLogsSubscribe   = "action.octant.dev/podLogs/subscribe"
 	RequestPodLogsUnsubscribe = "action.octant.dev/podLogs/unsubscribe"
+	DefaultSinceSeconds       = 300
 )
 
 type podLogsStateManager struct {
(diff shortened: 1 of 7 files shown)


internal/api/container_logs.go
```diff
@@ -81,23 +83,25 @@ func (s *podLogsStateManager) StreamPodLogsSubscribe(_ octant.State, payload act
...
 		return fmt.Errorf("getting since from payload: %w", err)
 	}
 
 	// Default to 5 minutes
-	sinceDuration := time.Second * 300
-	if since != 0 {
-		sinceDuration = time.Second * time.Duration(since)
+	sinceSeconds := int64(DefaultSinceSeconds)
+	// Allow negative since, negative since means since creation.
+	if since > 0 || since < 0 {
```
GuessWhoSamFoo, line 94: Isn't this still `since != 0`?

web/src/app/modules/shared/components/smart/logs/logs.component.ts
```diff
@@ -78,7 +78,7 @@ export class LogsComponent
   }
 
   onSinceChange(sinceSelection: number): void {
-    this.selectedSince = sinceSelection;
+    this.selectedSince = +sinceSelection;
```
GuessWhoSamFoo, line 81: Potentially redundant?

wwitzel3: I changed the `onSinceChange` handler to make this more clear. It now takes in the value as a string and converts it to a `number`. This is because the `(change)` $event.target.value will always be a string type.

GuessWhoSamFoo (approved):

PR #1438, opened 2020-10-01
use 127.0.0.1 for default DashService address

### LINKED ISSUE #1437: API instance for dashservice should assume 127.0.0.1 and not localhost

```
$ ./octant
unable to start runner: failed to start service api: initializing 
plugin manager: create dashboard api: create listener: listen 
tcp 198.105.254.23:0: bind: cannot assign requested address
```

octant started

https://github.com/vmware-tanzu/octant/blob/18f17d46b9000c4c1b413118ba7eb393e8d47837/pkg/plugin/api/api.go#L37-L48

This will do a DNS lookup for localhost. And if you're like me...your `/etc/hosts` will obviously be messed up because why wouldn't you mess with that file constantly :rofl:? Thus you'll get errors as can be seen above.

As stated in the net.Listen godoc:
```
// The address can use a host name, but this is not recommended,
// because it will create a listener for at most one of the host's IP
// addresses.
```
[xref](https://github.com/golang/go/blob/b8ec1d5f49ec5f9350f2b0bd99560e4aadfcb70c/src/net/dial.go#L694-L696)

**Environment:**

- Octant version (use `octant version`): `0.16.0`
- Kubernetes version (use `kubectl version`): `na`
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): `Linux taco 5.8.10-arch1-1 #1 SMP PREEMPT Thu, 17 Sep 2020 18:01:06 +0000 x86_64 GNU/Linux`

joshrosso:
Signed-off-by: joshrosso <rossoj@vmware.com>

This PR ensures the DashService API default to `127.0.0.1` rather than doing a DNS lookup against `localhost`.

- Fixes https://github.com/vmware-tanzu/octant/issues/1437

This fixes the issue and now is consistent with the listener address default:
https://github.com/vmware-tanzu/octant/blob/18f17d46b9000c4c1b413118ba7eb393e8d47837/internal/api/api.go#L32-L34

### DIFF
--- a/changelogs/unreleased/1438-joshrosso
+++ b/changelogs/unreleased/1438-joshrosso
@@ -1,0 +1,1 @@
+Default internal DashService API to 127.0.0.1 rather than localhost (DNS lookup)
--- a/pkg/plugin/api/api.go
+++ b/pkg/plugin/api/api.go
@@ -32,11 +32,13 @@ type grpcAPI struct {
 	listener net.Listener
 }
 
+const dashServiceAddress = "127.0.0.1:0"
+
 var _ API = (*grpcAPI)(nil)
 
 // New creates a new API instance for DashService.
 func New(service Service) (API, error) {
-	listener, err := net.Listen("tcp", "localhost:0")
+	listener, err := net.Listen("tcp", dashServiceAddress)
 	if err != nil {
 		return nil, errors.Wrap(err, "create listener")
 	}


GuessWhoSamFoo (approved):
Thanks @joshrosso !

I had to set localhost to something not 127.0.0.1 to replicate the error. Not sure how often this comes up,  but glad to avoid it entirely

PR #1463, opened 2020-10-06
fix markdown sanitization after upgrading to ngx-markdown 10

### LINKED ISSUE #1460: clarity icons stopped working in go plugin; is there alternative markup/markdown?
we've been using the lovely clarity icons inside markdown for things like ticks/crosses for indicating pipeline status. After upgrading to 0.16.x these get stripped out and do nothing now.

I wonder is there any way around this; or alternative markdown / markup we can use going forward? 

e.g. we were using markup like...

```html
<clr-icon shape="check-circle" class="is-solid is-success" title="Succeeded"></clr-icon>
or
<clr-icon shape="warning-standard" class="is-solid is-danger" title="Failed"></clr-icon>
```
though it looks like we can still use badges or spinners...

```html
<span class="spinner spinner-inline" title="Running"></span>
```

wwitzel3:

- Fixes #1460

### DIFF
--- a/changelogs/unreleased/1460-wwitzel3
+++ b/changelogs/unreleased/1460-wwitzel3
@@ -1,0 +1,1 @@
+Fix markdown being sanitized after upgrade to ngx-markdown 10
--- a/web/.storybook/preview.js
+++ b/web/.storybook/preview.js
@@ -3,12 +3,12 @@ import { SharedModule } from '../src/app/modules/shared/shared.module';
 import { MarkdownModule, MarkedOptions } from 'ngx-markdown';
 import { setConsoleOptions } from '@storybook/addon-console';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
-import { APP_INITIALIZER } from "@angular/core";
-import { InitService } from "../src/app/modules/shared/services/init/init.service";
-import { RouterModule } from "@angular/router";
-import { HttpClientModule } from "@angular/common/http";
-import { BrowserModule } from "@angular/platform-browser";
-import { AppRoutingModule } from "../src/app/app-routing.module";
+import { APP_INITIALIZER } from '@angular/core';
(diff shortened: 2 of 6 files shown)


scothis (dismissed):
Removing all sanitization can open up exploits when untrusted content is rendered as markdown. The exploit path boils down to:

1. create a resource with a reflected XSS backdoor in a value rendered as markdown
2. get an admin, or any other user with elevated privilege to view the resource in Octant
3. the reflected XSS executes in the admin's browser, effective performing operations against the cluster with their credentials, opening a backdoor into the cluster
4. use the installed backdoor to further exploit the cluster.

wwitzel3:
> Removing all sanitization can open up exploits when untrusted content is rendered as markdown. The exploit path boils down to:
> 
> 1. create a resource with a reflected XSS backdoor in a value rendered as markdown
> 2. get an admin, or any other user with elevated privilege to view the resource in Octant
> 3. the reflected XSS executes in the admin's browser, effective performing operations against the cluster with their credentials, opening a backdoor into the cluster
> 4. use the installed backdoor to further exploit the cluster.

I addressed the concerns above by explicitly setting the context to `SecurityContext.HTML` which is also the new default for ngx-markdown.

GuessWhoSamFoo (dismissed):
weird that linter didn't catch that for storybook

wwitzel3:
After re-testing with a fresh build, this does not fix the issue for `<clr-*>` tags working when using the markdown render.

wwitzel3:
This is because the default sanitizer strips out unknown HTML tags and it is not configurable, so we cannot extend it to allow tags explicitly.

jstrachan:
is there a way to use clarity icons using `<span class”...`>` syntax or something?

wwitzel3:
Sadly no, I've come up with another solution, but I'm not sure we can release it as a minor release, I'm closing this PR, but I'll put up the other one shortly.

Issue #1386, opened 2020-09-22
Events first/last seen can be null


GuessWhoSamFoo:
**What steps did you take and what happened:**
Currently this shows as `737689d`.

**What did you expect to happen:**
Show `<unknown>` or null.

**Environment:**

- Octant version (use `octant version`): 0.15

sladyn98:
@GuessWhoSamFoo would love to start on this. Any tips on how to begin

wwitzel3:
I think the place I would start exploring is https://github.com/vmware-tanzu/octant/blob/master/internal/printer/event.go

I'd add a test to https://github.com/vmware-tanzu/octant/blob/master/internal/printer/event_test.go that provides null values for the first/last seen, then work in event.go to convert that in to a display value instead like `<uknown>` of `737689d`

GuessWhoSamFoo:
Protip: use dev tools on chrome/firefox (ctrl + shift + i) > Network > stream. See what the server is returning for events in the json blob. The timestamp component converts a unix timestamp into a human readable format so there is likely some bad input causing an expected conversion.

PR #1527, opened 2020-10-20
Fix the first/last seen to display null instead of a pointer

sladyn98:

This PR tries to fix the first/last seen value to null instead of a pointer 
- Fixes #
https://github.com/vmware-tanzu/octant/issues/1386
This is a draft PR for now so feedback would help me move in the right direction
**Release note**:
```
release-note
```

### DIFF
--- a/internal/printer/event_test.go
+++ b/internal/printer/event_test.go
@@ -191,8 +191,6 @@ func Test_ReplicaSetEvents(t *testing.T) {
 					Component: "replicaset-controller",
 				},
 				Message:        "Created pod: frontend-b7fxg",
-				FirstTimestamp: metav1.Time{Time: time.Time{}},
-				LastTimestamp:  metav1.Time{Time: time.Time{}},
 			},
 		},
 	}
@@ -236,8 +234,8 @@ func Test_ReplicaSetEvents(t *testing.T) {
 		"Message":    component.NewText("Created pod: frontend-b7fxg"),
 		"Reason":     component.NewText("SuccessfulCreate"),
 		"Type":       component.NewText("Normal"),
-		"First Seen": component.NewText("Nil"),
-		"Last Seen":  component.NewText("Nil"),
+		"First Seen": component.NewText("<unknown>"),
+		"Last Seen":  component.NewText("<unknown>"),
 		"From":       component.NewText("replicaset-controller"),
 		"Count":      component.NewText("1"),
 	})
--- a/internal/printer/event_test.go
+++ b/internal/printer/event_test.go
@@ -176,6 +176,24 @@ func Test_ReplicaSetEvents(t *testing.T) {
 				FirstTimestamp: metav1.Time{Time: time.Unix(1548424410, 0)},
 				LastTimestamp:  metav1.Time{Time: time.Unix(1548424410, 0)},
 			},
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "frontend",
+					CreationTimestamp: metav1.Time{
+						Time: now,
+					},
+					ResourceVersion: "3",
+				},
+				Count:  1,
+				Type:   corev1.EventTypeNormal,
+				Reason: "SuccessfulCreate",
+				Source: corev1.EventSource{
+					Component: "replicaset-controller",
+				},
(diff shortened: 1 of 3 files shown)


pkg/view/component/timestamp.go
```diff
@@ -27,12 +27,21 @@ type TimestampConfig struct {
 
 // NewTimestamp creates a timestamp component
 func NewTimestamp(t time.Time) *Timestamp {
-	return &Timestamp{
-		Base: newBase(TypeTimestamp, nil),
-		Config: TimestampConfig{
-			Timestamp: t.Unix(),
-		},
+
+	actualTime, err := time.Parse(time.RFC1123 ,t.String())
```
sladyn98, line 31: Parsing the time would help us sanitize and see if the format is correct. Any other approaches are welcome

pkg/view/component/timestamp.go
```diff
@@ -27,12 +27,21 @@ type TimestampConfig struct {
...
+	if err != nil {
+		return &Timestamp{
+			Base: newBase(TypeTimestamp, nil),
+			Config: TimestampConfig{
+				Timestamp: t.Unix(),
+			},
+		}
+	} else {
+		
+		// Need to return something if it is not of proper format
```
sladyn98, line 42: Should we return a zero time here or a null value

sladyn98 (commented):
@GuessWhoSamFoo @wwitzel3

internal/printer/event_test.go
```diff
@@ -176,6 +176,24 @@ func Test_ReplicaSetEvents(t *testing.T) {
...
+					ResourceVersion: "3",
+				},
+				Count:  1,
+				Type:   corev1.EventTypeNormal,
+				Reason: "SuccessfulCreate",
+				Source: corev1.EventSource{
+					Component: "replicaset-controller",
+				},
+				Message:        "Created pod: frontend-b7fxg",
+				FirstTimestamp: metav1.Time{Time: time.Time{}},
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Since the struct field is optional, omit `FirstTimestamp`/`LastTimestamp` entirely. After the conversion to int64, it'll be some really large negative number. Passing this value into `relative.pipe.spec.ts` reproduces the problematic timestamp seen in the issue.

pkg/view/component/timestamp.go
GuessWhoSamFoo: Check https://github.com/vmware-tanzu/octant/blob/master/internal/printer/event.go#L65-L66 and `PrintEvents` for nil before creating the timestamps, otherwise use a text component to show they are missing.

pkg/view/component/timestamp.go
GuessWhoSamFoo: I wouldn't worry about this part as the behavior of timestamp component is as expected, and the issue is about not passing bad inputs.

internal/printer/event_test.go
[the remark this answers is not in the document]
sladyn98: Okay so I will omit it entirely.

pkg/view/component/timestamp.go
sladyn98: Understood :+1:

pkg/view/component/timestamp.go
sladyn98: Done :+1:

GuessWhoSamFoo:
The changes look mostly good. You'll need to resolve any linting issues with `go run build.go fmt` and fix the unit test which can also be run locally via `go run build.go test`

wwitzel3:
Thank you for this @sladyn98 , once you address the things @GuessWhoSamFoo mentioned, you will also need to do a DCO (Developer Certificate of Origin) you can find out how to do that here: https://github.com/vmware-tanzu/octant/pull/1527/checks?check_run_id=1297519769

wwitzel3:
Hey @sladyn98 wanted to check in and see how things are going? Anything you need from us?

sladyn98:
@wwitzel3  Thanks I was just caught up in some stuff. I cleaned up the tests and the linters.

GuessWhoSamFoo (approved):
lgtm, thanks @sladyn98 !

PR #1536, opened 2020-10-21
Fix for navigations issue for plugins that supply leading slash in Urls

### LINKED ISSUE #1535: Wrong navigation items are selected for deep paths in a plugin

When navigating to a deeply nested link managed by a plugin, the wrong navigation item is highlighted and the wrong sub nav is shown. I don't see this behavior with the built-in navigation items.

Paths like `/knative` and `/knative/serving`, show the correct navigation, but deeper paths like `/knative/serving/services` show the wrong navigation menu.

Screen Shot 2020-10-21 at 11 14 36 AM

The plugins navigation items and menu should be shown when viewing a page defined by that plugin

[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Octant version (use `octant version`): HEAD

mklanjsek:
Handle the Url detection correctly for plugins that provide Urls starting with leading `/`. Should fix problems latest Navigation update was having with `knative` plugin.    

- Fixes #1535


mklanjsek:
NOTE: need to rebase after #1530 is merged, this branch is ahead of upstream.

wwitzel3 (approved):

PR #1557, opened 2020-10-26
Display Ingress host as a link

### LINKED ISSUE #1431: Ingress hosts field should be a link
On the Ingress summary page, inside the table containing all known ingresses, the `Hosts` column should be clickable. Currently we have it as a label:

Screen Shot 2020-09-29 at 1 28 11 PM

mklanjsek:
Instead of showing plain text, we will now try to detect if ingress host field contains the Url, and show it as a link if it does. 

This change also modifies the Link component to open the **external** links inside the separate browser window. This makes sense especially with a move to Electron app. All internal Octant links are handled same as before.

- Fixes #1431

### DIFF
--- a/internal/printer/ingress.go
+++ b/internal/printer/ingress.go
@@ -40,9 +40,10 @@ func IngressListHandler(ctx context.Context, list *extv1beta1.IngressList, optio
 			return nil, err
 		}
 
+		isTLS := len(ingress.Spec.TLS) > 0
 		row["Name"] = nameLink
 		row["Labels"] = component.NewLabels(ingress.Labels)
-		row["Hosts"] = getHostComponent(formatIngressHosts(ingress.Spec.Rules))
+		row["Hosts"] = getHostComponent(formatIngressHosts(ingress.Spec.Rules), isTLS)
 		row["Address"] = component.NewText(loadBalancerStatusStringer(ingress.Status.LoadBalancer))
 		row["Ports"] = component.NewText(ports)
 		row["Age"] = component.NewTimestamp(ingress.CreationTimestamp.Time)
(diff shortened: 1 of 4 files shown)


web/src/app/modules/shared/components/presentation/link/link.component.html
```diff
@@ -1,5 +1,5 @@
 <ng-template [ngIf]="isExternal" [ngIfElse]="internal">
-  <a href="{{ ref }}">{{ value }}</a>
+  <a target ="_blank" href="{{ ref }}">{{ value }}</a>
```
wwitzel3, line 2: Let's make sure we test this with a plugin and make sure that changing the target here doesn't impact the expected behavior.

mklanjsek: Verified that both internal and external links work as expected inside my test plugin. Also tested that all links are working as expected inside existing plugins `knative` and `Jenkins X`

mklanjsek:
@wwitzel3 rebased and ready to go

internal/printer/ingress.go
```diff
@@ -54,6 +55,15 @@ func IngressListHandler(ctx context.Context, list *extv1beta1.IngressList, optio
 	return ot.ToComponent()
 }
 
+func getHostComponent(link string) component.Component {
+	_, err := url.Parse(link)
+	if strings.Contains(link, ".") && err == nil {
+		return component.NewLink("", link, "http://"+link)
```
wwitzel3, line 61: Can we look at the Ingress and determine if it is TLS/SSL?

wwitzel3: @mklanjsek ping

mklanjsek: Added TLS handling, also cases when multiple Urls are present.

mklanjsek:
@wwitzel3 PTAL

wwitzel3 (changes requested):
This is good, need to add some test cases to internal/printer/ingress_test.go 
There is already some decent test cases there to copy and extend.

mklanjsek:
@wwitzel3 good point, added tests around this functionality.

wwitzel3 (approved):

Issue #1504, opened 2020-10-15
Normalize form capabilities


scothis:
**Describe the problem/challenge you have**

The `form` component is used by the `modal` and `summary` components to render forms. The `stepper` component has its own form renderer that adds support for things like placeholders and validation.

**Describe the solution you'd like**

There should be a single system for defining forms so that experience can port between modals, summaries and steppers. 

**Environment:**

- Octant version (use `octant version`): 0.16.1

lenriquez:
I am working on this one

wwitzel3:
Currently we have forms embedded across many different components. For example the Modal, Stepper, and Cards all have form elements, but don't use a unified `FormComponent`.

Investigate what it would take to create a standard FormComponent and then use that component in the various places we want to render forms.

wwitzel3:
Created #2223 to track the creation of the form component which will then be used to normalize the behavior.

PR #1563, opened 2020-10-27
Refactor form payloads

### LINKED ISSUE #1508: Forms in modals produce invalid action.octant.dev/performAction events

When submitting a form from inside a modal component, the event structure nests the action payload so that the action cannot fire as expected.

Forms should work the same regardless of how they are rendered within the UI.

The event I expect to see is something like
```
{"type":"action.octant.dev/performAction","payload":{"action":"...",...}}
```

but I actually see

```
{"type":"action.octant.dev/performAction","payload":{"formGroup":{"action":"...",...}}}
```

**Environment:**

GuessWhoSamFoo:
This PR contains **breaking changes** to the plugin API around form usage. Since these patterns are not yet widely used (or unusable in the case of knative), changes can still be merged into 0.16.2

1. Check boxes
Previously check boxes with multiple input choices would generate a payload with only the checkbox label and a bool if any of the input choices were selected. e.g.
```
component.NewFormFieldCheckBox("checkbox", "test", []component.InputChoice{
	{
		Label:   "a",
		Value:   "1",
		Checked: true,
	},
	{
		Label:   "b",
		Value:   "2",
		Checked: false,
	},
})
```
Results in the loss of input choice values:
```
{
   "type":"action.octant.dev/performAction",
   "payload":{
      "test":true,
      "action":"action.octant.dev/example
   }
}
```
This is now changed to return an array of checked values similar to the behavior of `Select`.

2. Remove `formGroup`
Components were previoues using websocket service and sending messages with payloads. This is now moved to action service to deduplicate `action.octant.dev/performAction`. `formGroup` is removed so that plugin authors don't have to parse `request.Payload.Raw` and use `request.Payload.String` directly. ref #1504 

3. Marshal action to form
Forms added to modals will now be read by action handler in plugins.

- Fixes #1508 

### DIFF
--- a/changelogs/unreleased/1563-GuessWhoSamFoo
+++ b/changelogs/unreleased/1563-GuessWhoSamFoo
@@ -1,0 +1,1 @@
+Removed formGroup from payload actions and updated to return values for multiple checkboxes
--- a/pkg/view/component/form.go
+++ b/pkg/view/component/form.go
@@ -620,6 +620,7 @@ type Form struct {
 func (f *Form) MarshalJSON() ([]byte, error) {
 	t := struct {
 		Fields []map[string]interface{} `json:"fields"`
+		Action string                   `json:"action,omitempty"`
 	}{}
 
 	for _, field := range f.Fields {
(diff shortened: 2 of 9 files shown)


wwitzel3:
@scothis do you have a moment to try this out?

wwitzel3 (approved):

scothis:
@wwitzel3 seems to work https://github.com/vmware-tanzu/octant-plugin-for-knative/pull/75

Issue #1565, opened 2020-10-29
Daemonset incorrectly shows "Ok" with green check when # of Ready pods != Desired


bbgobie:
**What steps did you take and what happened:**
View from overall summary and Daemon Sets page is showing daemonset with green checkmark and "OK" status when hovered over.  However here is the output from Kubectl which Octant agrees with.

kube-proxy                 46        46        45      1            45          beta.kubernetes.io/os=linux   90d

**What did you expect to happen:**
A red X or something to clearly show that something is wrong with the daemonset.  For example Alertmanager is sending alerts, Octant if used as an operational dashboard needs to be able to clearly and concisely show which resources have issues.
**Anything else you would like to add:**
Ideally resources not ready default to the top of lists, eg pod list of this daemonset

**Environment:**

- Octant version (use `octant version`): Version:  0.16.1
Git commit:  8aebb34922f83894fb02ad393740e96ee1b3d8fe
Built:  2020-10-07T23:36:32Z
- Kubernetes version (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.9", GitCommit:"a34b0bfd73e1e6b618969df59dfd8bbfdd1e0711", GitTreeState:"clean", BuildDate:"2020-09-17T06:35:17Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Windows 10

PR #1566, opened 2020-10-30
Allow table sorting by multiple keys

GuessWhoSamFoo:
~~This PR adds sorting with multiple keys and sorts items with an error status by default in order to increase visibility of unhealthy cluster items.~~ This is a **breaking API change** as `table.Sort` is in `pkg`

Since datagrids are paginated, this results in a poor user experience if the status of an object changes while a user is looking at it. We will need to rethink if the goal should be filtering instead.

- ref #1565 

### DIFF
--- a/changelogs/unreleased/1566-GuessWhoSamFoo
+++ b/changelogs/unreleased/1566-GuessWhoSamFoo
@@ -1,0 +1,1 @@
+Added multi key table sort and added reverse method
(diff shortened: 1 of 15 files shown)


GuessWhoSamFoo:
there was discussion about how this might look more maintainable if we separated out the reverse param into its own method and have a sort that only requires one or more keys

GuessWhoSamFoo:
Added `table.Reverse()` and removed the boolean from sort

wwitzel3 (approved):

wwitzel3 (approved):
Please add a change log entry for this and lets document this in the storybook stories?

GuessWhoSamFoo:
@alexbrand Sorry this took a while to get merged!

PR #1584, opened 2020-11-03
Improve calculating navigation selection from Url

mklanjsek:
The algorithm that’s calculating selection from Url was not handling well situations when plugin paths are forming by adding on 2nd level navigation - everywhere else that’s done on 3rd level.

Reworked it, so now it ranks all known paths against the Url and one with deepest match wins. This insures that we will match the selection regardless how plugin authors decide to build their Url structure.


scothis:
Works well in my testing

wwitzel3 (approved):

PR #1676, opened 2020-11-20
refactor dash.Runner startup

jamieklassen:
**What this PR does / why we need it**: Adds two simple integration tests for the two main startup flows (with and without a kubeconfig)

- Fixes #

**Special notes for your reviewer**: These tests require the test runner process to be able to create a directory, create/delete temporary files, and open TCP listeners (high-numbered random ports via `127.0.0.1:0`) on the host.

**Release note**:
```
release-note
```

### DIFF
--- a/pkg/dash/dash.go
+++ b/pkg/dash/dash.go
@@ -106,10 +106,7 @@ func NewRunner(ctx context.Context, logger log.Logger, options Options) (*Runner
 	} else {
 		logger.Infof("no valid kube config found, initializing loading API")
 		// Initialize the API
-		apiService, apiErr = r.initLoadingAPI(ctx, logger, options)
-		if apiErr != nil {
-			return nil, fmt.Errorf("failed to start loading api: %w", err)
-		}
+		apiService = api.NewLoadingAPI(ctx, api.PathPrefix, r.actionManager, r.websocketClientManager, logger)
 	}
 
 	d, err := newDash(options.Listener, options.Namespace, options.FrontendURL, options.BrowserPath, apiService, pluginService, logger)
@@ -169,12 +166,6 @@ func (r *Runner) Start(ctx context.Context, logger log.Logger, options Options, 
 
 	shutdownCh <- true
 	return nil
-}
-
-func (r *Runner) initLoadingAPI(ctx context.Context, logger log.Logger, _ Options) (*api.LoadingAPI, error) {
-	apiService := api.NewLoadingAPI(ctx, api.PathPrefix, r.actionManager, r.websocketClientManager, logger)
-
-	return apiService, nil
 }
 
 func (r *Runner) initAPI(ctx context.Context, logger log.Logger, options Options) (*api.API, *pluginAPI.GRPCService, error) {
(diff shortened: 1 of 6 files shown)


jamieklassen:
Pushed some low-hanging refactors to clean up `Runner.Start` especially.

I've toyed with ways to break the filesystem dependency (even though an in-memory `afero.Fs` be used as a double for testing `ValidateKubeConfig`, `cluster.FromKubeConfig` still calls out to the real filesystem) or the use of a 'real' `httptest.Server` for the fake k8s API, but both would require really intrusive refactorings and result in duplicating some of the code from `k8s.io/client-go` so I think it's worth paying this performance/stability price in the tests.

However after discussing with @wwitzel3, tomorrow I will add a flag and a field on `dash.Options` to allow specifying the host:port for octant. This seems to be a sensible feature that nobody has yet gotten around to. Ideally this will also pave the way to use a a fake, in-memory `net.Listener` (not backed by a real network interface) in tests.

GuessWhoSamFoo:
I'll take a pass at this later today

pkg/dash/dash.go
```diff
@@ -129,39 +125,42 @@ func NewRunner(ctx context.Context, logger log.Logger, options Options) (*Runner
...
-				} else {
-					logger.Errorf("unexpected empty kube config")
-					return
-				}
-			}
-		}()
-	} else {
-		r.startAPIService(ctx, logger, options)
+		logger.Infof("waiting for kube config ...")
+		options.KubeConfig = <-ocontext.KubeConfigChFrom(r.ctx)
```
GuessWhoSamFoo, line 138: This blocks which is fine because it needs to wait if the user adds a kubeconfig, however it will hang on sigint/sigterm without shutting down plugins. In other words, octant will hang if the users tries to exit in the loading api state via `ctrl + c` without a way to recover

GuessWhoSamFoo (changes requested):
This looks good. It greatly improves the readability of `dash.go`

pkg/dash/dash.go
jamieklassen: oh that makes sense, thank you! I'll fix this up.

GuessWhoSamFoo (commented):
lgtm, can you squash some of these commits?

GuessWhoSamFoo (approved):

jamieklassen:
@GuessWhoSamFoo I rearranged the changes to remove some of the re-work -- for example, in the old commit log I removed the "wait for kubeconfig in a goroutine" logic and then re-added it based on your feedback, whereas in this version I add that integration test in the first commit and ensure that all commits pass the tests -- but I kept the individual commits for each stage of the refactor.

I thought it would be valuable to keep each change separate, along with its justification. This is mainly because the tests I added are not exhaustive, and I may have made another mistake besides the one you already observed. If that turns out to be the case, then somebody may need to look back at my changes. I imagine it would be helpful to have a log of each refactoring decision as it was made.

I'm OK to squash more aggressively if you prefer. Let me know! 😄

PR #1703, opened 2020-12-03
Add support for ephemeral containers

### LINKED ISSUE #1103: Create ephemeral containers for debugging
Some containers might not have bash or sh for a container exec.

An implementation of 
https://kubernetes.io/docs/tasks/debug-application-cluster/debug-running-pod/#ephemeral-container

This feature could have the following additions:
 - Failing to create a terminal when navigating to the tab prompts user to create an ephemeral container.
 - Modify object status to remind users to delete pods with a running ephemeral container.

n/a

**Environment:**

- Octant version (use `octant version`): 0.13.1
- Kubernetes version (use `kubectl version`): 1.18

GuessWhoSamFoo:
This PR contains an initial pass at supporting ephemeral containers in Octant. If the feature gate is enabled when visiting a pod, it will automatically add debug container using a debian image.

Pods with an ephemeral container will have their object status modified as `warning` (yellow).

If there is interest in this feature, further efforts can be invested into making the debug container configurable and proper UI.

- Fixes #1103 

Example of a distroless image where `sh` and `bash` will not work:

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
spec:
  selector:
    matchLabels:
      app: nginx
  replicas: 3
  template:
    metadata:
      labels:
        app: nginx
    spec:
      shareProcessNamespace: true
      containers:
      - name: nginx
        image: sfoo/nginx-distroless:latest
        ports:
        - containerPort: 80
        - containerPort: 443
```

Testing if a feature gate is enabled using mocked clients is not straightforward and I haven't figured out a way to do this yet. The terminal viewer component is pending a refactor soon because the terminal viewer does not have knowledge of the manager's state which causes unnecessary synchronization challenges.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -68,7 +68,9 @@ require (
 	k8s.io/api v0.19.3
 	k8s.io/apiextensions-apiserver v0.19.3
 	k8s.io/apimachinery v0.19.3
+	k8s.io/apiserver v0.19.3
 	k8s.io/client-go v0.19.3
+	k8s.io/component-base v0.19.3
 	k8s.io/klog v1.0.0
 	k8s.io/klog/v2 v2.3.0 // indirect
 	k8s.io/kube-aggregator v0.19.3
--- a/internal/api/terminal_manager_test.go
+++ b/internal/api/terminal_manager_test.go
@@ -1,0 +1,28 @@
+package api_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+
+	"github.com/vmware-tanzu/octant/internal/api"
+	"github.com/vmware-tanzu/octant/internal/api/fake"
+	configFake "github.com/vmware-tanzu/octant/internal/config/fake"
+	octantFake "github.com/vmware-tanzu/octant/internal/octant/fake"
+)
+
+func Test_TerminalStateManager(t *testing.T) {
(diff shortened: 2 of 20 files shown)


internal/api/terminal_manager.go
```diff
@@ -136,16 +136,24 @@ func (s *terminalStateManager) startStream(key store.Key, container string) cont
...
 	eventType := event.NewTerminalEventType(key.Namespace, key.Name, container)
 
-	instance, err := terminal.NewTerminalInstance(ctx, s.config.ClusterClient(), logger, key, container, "/bin/sh", s.chanInstance)
-	if err != nil {
-		cancelFn()
-		return cancelFn
-	}
-
-	s.instance = instance
+	commands := []string{"sh", "bash", "powershell", "cmd"}
```
wwitzel3, line 139: This should be it's own PR with a changelog entry.

wwitzel3 (changes requested):
This looks pretty good, part of this should be a PR that lands in upstream as its own change though.

internal/api/terminal_manager.go
GuessWhoSamFoo: Removed and tracked in #1749

wwitzel3 (approved):

PR #1721, opened 2020-12-10
Custom clientconfig

### LINKED ISSUE #1744: Allow data interfaces in Octant to be injected at runtime.
Octant has a handful of data interfaces for fetching various Kubernetes resources and performing work such as log streaming and establishing interactive terminals. Currently all of these interfaces are defined internally to Octant and instances of objects that satisfy these interfaces are created internally to the Octant process.

We want to create the ability to have another tools and projects that may wish to use the Octant stream generator and plugin system to be able to inject their own objects that satisfy these interfaces at runtime without the need to fork the project and edit things inside the `internals` package.

### First pass
Ensure all options can be configured at runtime using the `With` functional options pattern.

Then we will be to promote the interfaces we have in internal such as ClientInterface, NamespaceInterface, and InfoInterface in to `pkg` so that we can add a `WithOption`. Ultimately we'd want to have a single interface, like our `Store` interface that would abstract away the need for these implementation specific interfaces needing to also be satisfied at runtime.

### Caution
By promoting these interfaces to the `pkg` folder they become part of the public API of Octant. If this is only the case for a short period while we work to extend our Store interface to encapsulate all of the various fetching tasks, this isn't a problem. Long term, if these interfaces are left public for an extended period and have wide adoption, we would have more overhead in deprecating and migrating folks away from them.

mstergianis:
- Separates functionality of client interface and context switching
- Allows one to use client interfaces not backed by kubeconfigs

- Fixes no issues

**Release note**:
```
release-note
```

cc @dimitrybesson @zoetian


mstergianis:
The "Verify generated code" check appears to be failing because of [removing the Loader interface](https://github.com/vmware-tanzu/octant/pull/1721/files?file-filters%5B%5D=.go&hide-deleted-files=true#diff-2416f88d535bd940665a6d183c33508a77f267716ac70dd9526ca7bbf2c38b9dL17) as the changes we made caused the interface to become unecessary.

GuessWhoSamFoo:
Has the mock file been regenerated with `go run build.go go-install && go run build.go generate`?

mstergianis:
Good catch @GuessWhoSamFoo. I didn't pay enough attention to the diffs and didn't notice that the file was not deleted. I fixed the underlying issue: `removeFakes` in `build.go` was trying to remove `{pkg,internal}/<some package>/fake/fake`

wwitzel3 (changes requested):
Please clean up the commit history here. We want to avoid things like `WIP` in the mainline history if we can.

wwitzel3:
@GuessWhoSamFoo I created #1744 to add some much needed context around this change.

pkg/dash/watcher.go
```diff
@@ -132,7 +132,7 @@ func (cw *ConfigWatcher) Watch(ctx context.Context) {
 			done = true
 			logger.Infof("shutting down config watcher")
 		case _ = <-cw.FileWatcher.Events():
-			if err := cw.watcherConfig.UseContext(ctx, cw.watcherConfig.ContextName()); err != nil {
+			if err := cw.watcherConfig.UseContext(ctx, cw.watcherConfig.CurrentContext()); err != nil {
```
GuessWhoSamFoo, line 135: A potentially interesting edge case (doesn't have to be addressed here) is what should happen if a config is loaded but becomes invalid (e.g. deleted) while Octant is running.

internal/event/context.go
```diff
@@ -27,18 +24,21 @@ type kubeContextsResponse struct {
...
 }
 
 var _ octant.Generator = (*ContextsGenerator)(nil)
 
-func NewContextsGenerator(dashConfig config.Dash, options ...ContextGeneratorOption) *ContextsGenerator {
+func NewContextsGenerator(kubeContextStore KubeContextStore, options ...ContextGeneratorOption) *ContextsGenerator {
 	kcg := &ContextsGenerator{
-		ConfigLoader: kubeconfig.NewFSLoader(),
-		DashConfig:   dashConfig,
+		KubeContextStore: kubeContextStore,
```
GuessWhoSamFoo, line 41: Make sure the context is updating as the kubeconfig changes from the watcher. Current behavior has the context updating in Octant via `kubectl config use-context`. A similar case also holds if additional contexts are added though tools like kind.

Additionally, an instance of Octant running via `FromClientConfig` should not be looking at the filesystem.

GuessWhoSamFoo:
I haven't looked into why - these changes also resurface an older issue when switching contexts
```
2020-12-17T13:43:45.562-0800    ERROR   api/content_manager.go:154      generate content     {"client-id": "ccdafa4b-40b0-11eb-a4ed-
8c16455fb1b6", "err": "generate content: preferred resources: unable to retrieve the complete list of server APIs: 
metrics.k8s.io/v1beta1: the server is currently unable to handle the request", "content-path": "overview/namespace/kube-system"
```

mstergianis (commented):
> I haven't looked into why - these changes also resurface an older issue when switching contexts
> 
> ```
> 2020-12-17T13:43:45.562-0800    ERROR   api/content_manager.go:154      generate content     {"client-id": "ccdafa4b-40b0-11eb-a4ed-
> 8c16455fb1b6", "err": "generate content: preferred resources: unable to retrieve the complete list of server APIs: 
> metrics.k8s.io/v1beta1: the server is currently unable to handle the request", "content-path": "overview/namespace/kube-system"
> ```
> 
> EDIT: I actually do; we need to still generate content even if not in the list of preferred server resource.

I'm looking into this now, but am unable to reproduce by switching contexts in the UI or via `kubectl`. Are there any other steps needed to reproduce?

mklanjsek:
> Are there any other steps needed to reproduce?

@mstergianis To repro that issue, you need a failing resource on your cluster  - in this case `metrics.k8s.io/v1beta1` had missing endpoints.  That could be hard to generate, in that case maybe @wwitzel3 can give you access to our dev cluster.

mstergianis:
Thanks @mklanjsek, will do

wwitzel3:
The metrics issue was not re-introduced in this PR, it was re-introduced here: https://github.com/vmware-tanzu/octant/commit/dcb101861ff11cd521317464d29d81ed1b9047eb

wwitzel3:
Looks like the current failing test on here is a legitimate one? @mstergianis

mstergianis:
Okay, so I've moved off of this track of work, but have brought @xtreme-vikram-yadav and @weresch up to speed on why the tests are failing.

Before the tests can be fixed though I want to confirm what the intended behaviour is for octant to monitor filesystem changes to the kubeconfig.

Before PR:
1. octant starts up
2. octant begins monitoring user's kubeconfig in https://github.com/vmware-tanzu/octant/blob/master/pkg/dash/watcher.go#L126-L142
3. when the user switches context in octant via the websocket, a non empty string is written to this in memory field https://github.com/vmware-tanzu/octant/blob/915279cbe3431884a00c50a85b87b597e750a425/internal/config/dash.go#L122
4. octant no longer switches context when the user changes their kubeconfig on fs because the watcher passes in the current octant context in to the `UseContext` method

After PR:
1. octant starts up
2. octant begins monitoring user's kubeconfig in https://github.com/mstergianis/octant/blob/custom-clientconfig/pkg/dash/watcher.go#L126-L142
3. when the user switches contexts via websocket, the watcher continues to pass the empty string into `UseContext` so the kubeconfig is always parsed, and the octant context is always synced with the filesystem

In light of [this](https://github.com/vmware-tanzu/octant/issues/1793) conversation with @GuessWhoSamFoo, we are unsure if the behaviour to ignore changes to the filesystem are intentional or not. This impacts if we fix the tests or the code.

wwitzel3:
We do not want Octant to persist context switching changes to the filesystem.

mstergianis:
Spoke to Wayne offline, we're going to modify the `KubeContextDecorator` so it behaves as closely as possible to the way Live switched contexts in the past.

wwitzel3:
Do you have any more questions about the behavior or the failing test?

wwitzel3:
How's it going on this?

weresch:
Hi Wayne, I've picked up this PR in the meantime. I've traced the behaviour of the original code to see how kube context is passed around. Now tracing the behaviour of the PR code so it can be updated to replicate the behaviour of the original code.

wwitzel3 (approved):

PR #1737, opened 2020-12-14
Only show preferred version of custom resources

### LINKED ISSUE #1715: Duplicate resources for CRDs with multiple conversionReviewVersions
I have a lot of custom resources in my cluster from Knative, and a lot of them are duplicated under both ...knative.dev/v1 and ...knative.dev/v1beta. I'm guessing this is due to the CRDs specifing multiple conversionReviewVersions, but it's a bit odd. (It would probably be a lot more aggravating if I had more of these resources.)

What's more frustrating is trying to edit these resources. If I try to edit any of the copies of the resource that isn't the most recent version, I get an error that the api version was not the expected one.

Would it be feasible to either indicate that the most recent version is the only editable and "correct" version, or hide the old versions completely?

**Environment:**
- Octant version (use `octant version`): Git commit:  06f530e0
- Kubernetes version (use `kubectl version`): v1.19.4
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Debian testing

bryanl:

When a CRD has multiple defined versions, they are all shown. This is
distracting and confusing for Octant users. Instead, only show the
preferred version, and if multiple versions exist, show them in the
summary for a custom resource.

Screen Shot 2020-12-14 at 3 15 05 PM
Screen Shot 2020-12-14 at 3 15 20 PM

- Fixes #1715

**Release note**:
```
Only show preferred version of custom resources```


wwitzel3 (approved):

wwitzel3:
This looks good to me, @GuessWhoSamFoo if you have a moment to try it out, I'd like your feedback here as well.

wwitzel3 (changes requested):
Sorry, I missed a thing, can you add a changelog entry for this. Thank you.

GuessWhoSamFoo:
Something weird happened on the rebase, but the core changes look fine to me.

wwitzel3 (approved):

PR #1746, opened 2020-12-17
Cluster interface refactor

### LINKED ISSUE #1744: Allow data interfaces in Octant to be injected at runtime.

mstergianis:
**What this PR does / why we need it**: Makes octant more flexible as a library. See #1721 for more context.

- In tandem with #1721 this fixes #1744 

I left `ClientInterface` in `internal` as it was not absolutely necessary to move to a public folder (something that was mentioned in #1744). Let me know if it is preferred to make that interface public, but because concrete implementations only need to implicitly implement the interface, they never need to explicitly reference it.

**Release note**:
```
release-note
```


mstergianis:
I left `ClientInterface` in `internal` as it was not absolutely necessary to move to a public folder (something that was mentioned in #1744). Let me know if it is preferred to make that interface public, but because concrete implementations only need to implicitly implement the interface, they never need to explicitly reference it.

wwitzel3 (approved):

Issue #944, opened 2020-05-20
Octant running an electron app should use random port for backend


bryanl:
Octant uses tcp://localhost:7777 as the default port for the backend listener. When running inside of the electron app, it should use a random port instead.

PR #1747, opened 2020-12-18
Allow selecting a random port

yeahdongcn:
Signed-off-by: yeahdongcn <yeahdongcn@gmail.com>

When running octant without specifying `OCTANT_LISTENER_ADDR ` or `--listener-addr`, octant sticks to port 7777. It reports `dial: already in use` then quits if more than one contexts are used. This PR allows the end-users to select a random port available after 7777 or whatever they previously specified.

`octant --context kind-kind` and `octant --context kind-xyz` can run successfully without manually specifying a different port.

- Unknown.

**Testing done**:
`go run build.go ci` passed.

**Release note**:
```
Octant now allows selecting a random port.
```

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -9,8 +9,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"math"
 	"net"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/api/api.go
```diff
@@ -57,11 +59,26 @@ func acceptedHosts() []string {
...
-		return net.Listen("tcp", listenerAddr)
+		panic(fmt.Sprintf("Unable to parse OCTANT_LISTENER_ADDR: %s", listenerAddr))
+	}
+	portNumber, err := strconv.Atoi(port)
+	if err != nil {
+		panic(fmt.Sprintf("Unable to convert port: %s", port))
+	}
+	for i := 0; ; i++ {
+		port := portNumber + i
+		listenerAddr = net.JoinHostPort(host, fmt.Sprintf("%d", port))
```
wwitzel3, line 72: If `OCTANT_LISTENER_ADDR` or `--listener-addr` is explicitly provided, we want to fail. We don't want to randomize ports when explicitly given a value.

wwitzel3:
First, thank you for this PR, really excited to see that you are using Octant and wanting to increase the value you get from it by contributing, thanks again.

For the PR itself, I don't think the approach here is the one we ultimately want to use with Octant. This would seem like a good place to make this change, but this is a result of the Octant launcher and the Octant server currently being baked in to the same executable. We are currently in the process of moving away from that model.

- We want the octant server to startup in a deterministic way.
- We want the octant server to always respect the value of OCTANT_LISTENER_ADDR or --listener-addr and fail if unable to bind.
- We need a way to capture the port that the Octant server ends running on, so that the Electron launcher and application can communicate with it.

If you are interested in contributing this change to the next release, you can look at our `main.ts` file in our `electron-builder` branch. This is where we will be finding the random port and ensuring that it is injected in to the all of the correct places. Issue #944 is the tracking issue. And here is a link to starting point for this change: https://github.com/vmware-tanzu/octant/blob/electron-builder/web/main.ts#L78

Issue #1696, opened 2020-11-30
Breadcrumb navigation not always working


frightener:
Hello,

the breadcrumb navigation doesn't seem to be working in all cases.
I'm currently viewing the node-exporter Daemon Set and the navigation shows: Workloads -> Daemon Sets -> node-exporter.

The Workloads link is correct and works: http://127.0.0.1:7777/#/overview/namespace/openshift-monitoring/workloads

The Daemon Sets link on the other hand seems broken: http://127.0.0.1:7777/#/overview/namespace/openshift-monitoring/workloads/daemon-sets/node-exporter/%5Coverview%5Cnamespace%5Copenshift-monitoring%5Cworkloads%5Cdaemon-sets

Regards,

Christian

GuessWhoSamFoo:
Hi Christian! Thanks for the bug report. It looks like a malformed link with bad url encoding as well. Could you share more information about where that link came from?

The only places I can find are the datagrid listing, resource viewer, and `controlled by` in the metadata tab from one of the pods which appear okay on inspection in 0.16.2

frightener:
I can see the malformed link anytime there is a navigation path with two components.

I click on Workload->Pods and then on one of the pods that are available in the current namespace. Above the pod name I have the following navigation path:

Workloads > Pods > node-exporter-7zhqr

"Workloads" and "Pods" are links, but only the first one is correct and the second one is malformed. This is true for other views as well (e.g. Workloads->Daemon Sets).

I'm using 0.16.2 as well.

mklanjsek:
Hey @frightener, I started looking into this but have hard time reproducing it in my setup. What is the OS and browser you are using? Looks like you were using OpenShift in above example - do you see the issue for deployments not generated by OpenShift? Anything that can help narrow down specific environment detail triggering the issue would be helpful.

What appears to be happening is that the breacrumb urls contain additional copy of resource path, where path is for some reason separated by backslash and encoded, so path separators show up as `%5c`.

lgellrich:
Hi @mklanjsek ,

I'm actually observing the very same issue only on Windows, while a Linux client with the same octant version and Kubernetes cluster works as expected.

Looking at the Websocket traffic on Windows I do receive the following message:
```
{
  "type": "event.octant.dev/content",
  "data": {
    "content": {
      "title": [
        {
          "config": {
            "value": "Workloads",
            "ref": "/overview/namespace/kube-system/workloads"
          },
          "metadata": {
            "type": "link",
[...]

PR #1796, opened 2021-01-07
Fixed issues with incorrect urls on Windows

mklanjsek:
In some cases when doing url string manipulations, we were using `filepath` instead of `path` go package.  Using OS specific `filepath` package on Windows resulted in incorrect urls that contain backslashes, as noted in #1696.

The original problem from #1696 is not present any more because that code is eliminated with latest breadcrumbs improvements. However, I found another occurrence where `filepath` was used when generating navigation entries and fixed it in this PR.

Big thanks to @lgellrich for correctly identifying the problem.

- Fixes #1996

### DIFF
--- a/changelogs/unreleased/1696-mklanjsek
+++ b/changelogs/unreleased/1696-mklanjsek
@@ -1,0 +1,1 @@
+Fixed issues with incorrect paths on Windows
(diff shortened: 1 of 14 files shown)


GuessWhoSamFoo:
I'd like to see a test for this across all modules or moving this to a common helper to generate root paths.

mklanjsek:
thanks @GuessWhoSamFoo , decided to create path_util package and refactored code to use it.

GuessWhoSamFoo (approved):

Issue #1480, opened 2020-10-09
add electron asset building to our github actions


wwitzel3:
Ensure that our nightly and release pipelines are building the electron application.

PR #1827, opened 2021-01-15
Add electron build pipelines for nightly and release

### LINKED ISSUE #1479: add build-electron-prod to build.go
Add a new command to build.go to build a production binary of electron. This binary would bundle the API and start it automatically at runtime, unlike the binary built with `build-electron-dev`.

GuessWhoSamFoo:
- Fixes #1479, #1480

Additional changes (in future PRs) which should follow:

 - Switch from using goreleaser containers in asset building to `goreleaser/goreleaser-action@v2`. The action now allows an install-only mode to handle any pre-processing in the same runner. This can skip the upload/download steps for bundled assets files.
 - Wait for glob support on `actions/upload-release-asset@v1`. Although goreleaser supports uploading arbitrary files (https://github.com/goreleaser/goreleaser/pull/1333), it needs post processing of `checksums.txt` before upload which is not supported.
 - Ensure octant node subprocess can access local filesystem/env vars. The octant subprocess in electron does not have access to the local filesystem or environment variables and asks for a kubeconfig each time.
- Synchronize version across front and backend code. `electron-builder` enforces semver in `package.json` so version bumps need to happen in two places. Documentation and/or automation is needed to avoid potential mismatch. 
- Investigate high CPU usage. Something about running octant in this mode is causing CPU spikes at launch.
- Electron nightlies are relatively large files so the current behavior is to overwrite old uploads.
- Add a build flag to exclude front end bundles from `go build ...`.


web/main.ts
```diff
@@ -57,11 +58,51 @@ function createWindow(): BrowserWindow {
...
+  if (local) {
+    serverBinary = path.join(__dirname, 'extraResources', 'octant');
+  } else {
+    serverBinary = path.join(process.resourcesPath, 'extraResources', 'octant');
+  }
+
+  const server = child_process.spawn(serverBinary, [], {
+    env: { NODE_ENV: 'production', PATH: process.env.PATH },
+    detached: true,
+    stdio: ['ignore', out, err],
```
wwitzel3, line 82:
```
const server = child_process.spawn(serverBinary, ['--disable-open-browser'], { // don't try to open the browser
    env: { ...process.env, NODE_ENV: 'production' }  // this makes sure all the environment variables are passed in
```

wwitzel3 (changes requested):
This is looking great, one minor fix that will also remove a TODO item and this is ready to merge.

web/main.ts
GuessWhoSamFoo: Updated to pass flag and system env vars

wwitzel3 (approved):

PR #1863, opened 2021-01-25
Add resource viewer to storybook

GuessWhoSamFoo:
@ibagha JSON for a resource viewer can be previewed through the storybook

Some notes:
 - Color mismatch when resource is in warning
image
- Same panel above does not auto update with new data
- There is no visual difference between implicit and explicit connections
- Storybook only: different subcomponents sometimes don't show up. e.g. adding a label component to a node and text component to another results in only one visible at time


GuessWhoSamFoo:
Closing this in favor of using local content instead.

PR #1864, opened 2021-01-25
Donut chart improvements

### LINKED ISSUE #1465: Expose colour and tooltip properties for each segment in the DonutChart component

In starboard-octant-plugin, which is written in Go, we'd like to use the DonutChart component to visualise a vulnerability summary. However, we cannot map predefined colours (red, orange, green) to all severity levels (CRITICAL,HIGH,MEDIUM,LOW,UNKNOWN)

octant-plugin

Make the DonutChart component more generic so it can represent Pod or Node status, as well as such custom thing such as vulnerability summaries.

N/A

**Environment:**

mklanjsek:
This change improves donut chart interactivity and adds following new functionality:

- Donut chart now supports custom colors on top of existing support for predefined types,
- Added thickness property for whole chart as well as individual segments,
- Improved hover indication and added tooltip for selected segment,
- Fixed chart sizing and some other small issues,

- Fixes #1465

### DIFF
--- a/pkg/view/component/donut_chart.go
+++ b/pkg/view/component/donut_chart.go
@@ -20,14 +20,18 @@ type DonutChartLabels struct {
 }
 
 type DonutSegment struct {
-	Count  int        `json:"count"`
-	Status NodeStatus `json:"status"`
+	Count       int        `json:"count"`
+	Status      NodeStatus `json:"status"`
+	Color       string     `json:"color,omitempty"`
+	Description string     `json:"description,omitempty"`
+	Thickness   int        `json:"thickness,omitempty"`
 }
 
 type DonutChartConfig struct {
-	Segments []DonutSegment   `json:"segments"`
-	Labels   DonutChartLabels `json:"labels"`
-	Size     DonutChartSize   `json:"size"`
+	Segments  []DonutSegment   `json:"segments"`
+	Labels    DonutChartLabels `json:"labels"`
+	Size      DonutChartSize   `json:"size"`
+	Thickness int              `json:"thickness,omitempty"`
 }
 
 // +octant:component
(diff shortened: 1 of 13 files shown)


mklanjsek:
Note that for single segment charts we show small gap at the end of the segment. This is due to limitation of svg arc functionality and I decided not to work around it because it's consistent with the way we show gaps between segments. Could be addressed in future if deemed necessary.

web/src/app/modules/shared/components/presentation/donut-chart/donut-chart.component.ts
```diff
@@ -73,25 +75,132 @@ export class DonutChartComponent extends AbstractViewComponent<DonutChartView> {
...
+        this.tooltipX = `${event.offsetX}px`;
+        this.tooltipY = `${
+          event.offsetY -
+          event.target.ownerSVGElement.clientHeight -
+          scrollOffset -
+          48
+        }px`;
+        this.tooltipText = segment.description
+          ? segment.description
+          : `${segment.count} ${label} with status ${segment.status}`;
```
GuessWhoSamFoo, line 163: Node status can still return as an empty string if not set

pkg/view/component/donut_chart.go
```diff
@@ -63,6 +67,10 @@ func (dc *DonutChart) SetSize(size DonutChartSize) {
 	dc.Config.Size = size
 }
 
+func (dc *DonutChart) SetThickness(thickness int) {
+	dc.Config.Thickness = thickness
```
GuessWhoSamFoo, line 71: As a plugin developer, how will I know what is a reasonable number for thickness?

Negative thickness also makes interesting charts, but maybe it shouldn't be allowed
image

GuessWhoSamFoo (changes requested):
The current model doesn't lend well to generalized use cases of donut charts (e.g. showing percentages). Not a big deal here, but something to think about as other graphs are implemented

web/src/app/modules/shared/components/presentation/donut-chart/donut-chart.component.ts
mklanjsek: Good point, made a change to omit the status if not available.

pkg/view/component/donut_chart.go
mklanjsek: Right. Trimming this to [2-100] now. Also added comment to `SetThickness` method and story.

mklanjsek:
@GuessWhoSamFoo PTAL

GuessWhoSamFoo (approved):

PR #1956, opened 2021-02-05
Enable dynamic resisizing of header contents

GuessWhoSamFoo:
This PR allows dynamic resizing of the input filter based on viewport width. Additionally, media queries are used to hide extraneous items to prevent flex wrapping. Although the minimum window size is 400x400px, this method works until 600px (any smaller requires a different design approach).

image

image

cc @ibagha 


wwitzel3 (approved):
This looks good to me!

PR #1986, opened 2021-02-10
Developer Preferences

### LINKED ISSUE #1923: Use electron-store for preferences and consolidate storage.
#1906 introduces electron-store.
#1538 adds preference UI.
#941 introduced server side preference `Frontend Proxy URL`
#1885 uses a `state.json` file to store the last known x,y,w,h for the application window.

Let's consolidate these storage locations. I think storing everything using `electron-store` makes a lot of sense here.

I think due to some points raised in #1538 this work should happen before we release 0.17.

cc: @GuessWhoSamFoo @mklanjsek

mklanjsek:
This change adds the Developer preferences, available only in Electron build. More precisely, it adds developers ability to change the logging verbosity and specify the Frontend Url.

Additional info:
- Also fixed small css issues with `Preferences` button,
- Fixed problem where validation was not running on Preferences form load,

- Fixes #1923

### DIFF
--- a/web/main.ts
+++ b/web/main.ts
@@ -32,8 +32,6 @@ let tray = null;
 
 const args = process.argv.slice(1);
 const local = args.some(val => val === '--local');
-const defaultRetries = 5;
-let numberRetries= defaultRetries;
 
 const applicationMenu = new ApplicationMenu();
 Menu.setApplicationMenu(applicationMenu.menu);
@@ -48,18 +46,8 @@ function saveBoundsSoon() {
   }, 1000);
 }
 
-function loadFrontend() {
-  numberRetries--;
-  if (electronStore.get('development').embedded || numberRetries < 0) {
-    if(numberRetries < 0) {
-      const alertOptions: MessageBoxOptions = {
-        type: 'warning',
-        buttons: ['OK'],
-        message: `Reverted to Embedded Frontend because the Frontend proxy Url is not reachable.`,
-        detail: 'Make sure the front end service is running on Frontend proxy Url.',
-      };
-      dialog.showMessageBoxSync(win, alertOptions);
-    }
+function loadFrontend(embedded:boolean) {
+  if (embedded) {
     win.loadFile(path.join(__dirname, 'dist/octant/index.html'));
   } else {
     win.loadURL(electronStore.get('development').frontendUrl);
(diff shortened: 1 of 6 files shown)


web/main.ts
```diff
@@ -45,6 +48,24 @@ function saveBoundsSoon() {
   }, 1000);
 }
 
+function loadFrontend() {
+  numberRetries--;
```
wwitzel3, line 52: I think we should remove this retry logic. It doesn't provide any value in practice since it is a tight loop. It would be better if we could give the user the option to retry when ready, this lets them (me) go run `npm run start` in cases when I had forgot

web/main.ts
```diff
@@ -45,6 +48,24 @@ function saveBoundsSoon() {
...
+  if (electronStore.get('development').embedded || numberRetries < 0) {
+    if(numberRetries < 0) {
+      const alertOptions: MessageBoxOptions = {
+        type: 'warning',
+        buttons: ['OK'],
+        message: `Reverted to Embedded Frontend because the Frontend proxy Url is not reachable.`,
```
wwitzel3, line 58: `Frontend Proxy URL is unreachable.`

web/main.ts
```diff
@@ -45,6 +48,24 @@ function saveBoundsSoon() {
...
+        detail: 'Make sure the front end service is running on Frontend proxy Url.',
```
wwitzel3, line 59: `Please ensure the frontend service is running at $<insert URL>.`

web/main.ts
```diff
@@ -45,6 +48,24 @@ function saveBoundsSoon() {
   }, 1000);
 }
 
+function loadFrontend() {
+  numberRetries--;
+  if (electronStore.get('development').embedded || numberRetries < 0) {
+    if(numberRetries < 0) {
+      const alertOptions: MessageBoxOptions = {
+        type: 'warning',
+        buttons: ['OK'],
```
wwitzel3, line 57: Can we make this two buttons, `Retry` and `Cancel`

Where retry would attempt to reload the proxy URL (essentially running this method again)
And Cancel would reset to the internal assets.

web/src/app/modules/shared/services/preferences/preferences.service.ts
```diff
@@ -92,6 +128,10 @@ export class PreferencesService implements OnDestroy {
     const collapsed = update['general.navigation'] === 'collapsed';
     const showLabels = update['general.labels'] === 'show';
     const isLightTheme = update['general.theme'] === 'light';
+    const frontendUrl = update['development.frontendUrl'];
+    const verbose = update['development.verbose'] === 'high';
```
wwitzel3, line 132: debug instead of high

web/src/app/modules/shared/services/preferences/preferences.service.ts
```diff
@@ -104,83 +144,173 @@ export class PreferencesService implements OnDestroy {
...
+            },
+          ],
+        },
+        {
+          name: 'Logging verbosity (requires restart)',
+          elements: [
+            {
+              name: 'development.verbose',
+              type: 'radio',
+              value: this.verbose.value ? 'high' : 'low',
```
wwitzel3, line 227: normal / debug for these and the labels.

wwitzel3 (changes requested):
This looks great. I had a couple minor label changes and a question about retry/alert.

web/main.ts
mklanjsek: Yeah, makes sense - removed.

web/main.ts
mklanjsek: ✅

web/main.ts
mklanjsek: ✅

web/main.ts
mklanjsek: ✅

web/src/app/modules/shared/services/preferences/preferences.service.ts
mklanjsek: Changed to `debug`/`normal` wording.

web/src/app/modules/shared/services/preferences/preferences.service.ts
mklanjsek: ✅

wwitzel3 (approved):

PR #2047, opened 2021-02-19
Added table page size preference

mklanjsek:
- Added a new preference to set table page size with default value of 10,
- Added Reset button so the preferences can be restored to default state,
- Rearranged the panels - moved the navigation preferences to separate panel, 
- Refactored preferences handling to make it easier to add new and maintain existing 

Luis and Milan

### DIFF
--- a/web/src/app/modules/shared/components/presentation/preferences/preferences.component.html
+++ b/web/src/app/modules/shared/components/presentation/preferences/preferences.component.html
@@ -60,6 +60,11 @@
                         </div>
                       </div>
                     </ng-container>
+                    <ng-container *ngSwitchCase="'dropdown'">
+                      <label>{{ element.label }}</label>
+                      <app-view-dropdown [view]="element" (selectedValue)="onDropDownValueChange($event, element.name)">
+                      </app-view-dropdown>
+                    </ng-container>
                     <div *ngSwitchDefault>
                       <p>not sure what to do with {{ element.type }}</p>
                     </div>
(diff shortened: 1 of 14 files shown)


web/src/app/modules/shared/components/presentation/preferences/preferences.component.html
```diff
@@ -60,6 +60,11 @@ <h4>{{ section.name }}</h4>
                         </div>
                       </div>
                     </ng-container>
+                    <ng-container *ngSwitchCase="'dropdown'">
+                      <label>{{ element.label }}</label>
+                      <app-view-dropdown [view]="element" (selectedValue)="onDropDownValueChange($event, element.name)">
```
GuessWhoSamFoo, line 65: Why is dropdown used here over select (https://clarity.design/angular-components/select/)?

lenriquez: This is a good question, we are not using clarity here, I saw this component and just used

mklanjsek: I am OK with using our dropdown here, there a lots of benefits of dogfooding.

GuessWhoSamFoo: Implicitly this means using dropdown over select for all of preferences in order to maintain a consistent design. Are we okay with that?

wwitzel3: I think we should be consistent within the Clarity guidelines.

https://clarity.design/angular-components/dropdown/#dropdown-vs-select

mklanjsek:
Clarity selector looks good to me, thanks @lenriquez!

@GuessWhoSamFoo @wwitzel3 PTAL

GuessWhoSamFoo (approved):
lgtm

PR #2067, opened 2021-02-23
Reduce tooltip size on label to avoid hidden tags on K8s object

### LINKED ISSUE #2044: Unable to see all tags for a k8s object
**Issue**
When a component/object has more tags cannot see all the tags, only 7 - 8 logs are visible on window.

https://user-images.githubusercontent.com/51317257/108492309-01429380-72cb-11eb-9fce-fb89dea87ff8.mov

**Solution Required*
Add a vertical scrolling option so that tags above that window is visible

**Environment:**

- Octant version: 0.16.3
- Kubernetes version: 1.17.
- OS: Ubuntu 18.04

lenriquez:
Signed-off-by: Luis Enriquez <luisenriquez@vmware.com>

**What this PR does / why we need it**: Is not posible to see all the tags on k8s object 

- Fixes #2044

**Solution 1**
The tooltip for the labels does not use "position: fixed" but "position: absolute" and since the datagrid also use "position: absolute" part of the tooltip hides since is bigger than its parent.

```
::ng-deep .datagrid-outer-wrapper, 
::ng-deep .datagrid-inner-wrapper,
::ng-deep .datagrid {
  // Allow tooltip to be shown completely event if translate push them outside 
  // the grid
  overflow-y: visible;
}
```
**Problem with solution 1** 
This solution could work but according to this article 
https://css-tricks.com/popping-hidden-overflow/
is not posible some combinations like overflow-y: visible; and overflow-x: auto; so it does not looks well on small screens

**Solution 2** 
so the size of the tooltip was reduce to make sure it fit the table

**Problem with solution 2:** It looks too small

image

**Solution 3**
Found on https://github.com/vmware/clarity/issues/4891
remove position: relative from class datagrid-table

image

**Release note**:
```
release-note
```

### DIFF
--- a/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.scss
+++ b/web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.scss
@@ -43,9 +43,14 @@
     padding: 0 0.2rem;
   }
 
+  ::ng-deep &-content {
+    max-width: min-content;
+  }
+
   // To avoid horizontal scroll-bar inside of the signpost content body.
   ::ng-deep .signpost-wrap .signpost-content-body {
     padding: 1rem;
+    max-height: 12.2rem;
 
     .label:not(:last-child) {
       margin-bottom: 0.2rem;
--- a/changelogs/unreleased/2044-lenriquez
+++ b/changelogs/unreleased/2044-lenriquez
@@ -1,0 +1,1 @@
+Reduce tooltip size on label to avoid hidden tags on K8s object
--- a/web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.scss
+++ b/web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.scss
@@ -17,3 +17,7 @@
 app-button-group {
   float: right;
 }
+
+::ng-deep .datagrid-table {
+  position: static;
+}


GuessWhoSamFoo:
Long term - this needs to be fixed by Clarity (https://github.com/vmware-tanzu/octant/issues/2044#issuecomment-785554772).

I wonder if this can be temporarily patched by shifting the positions (https://clarity.design/angular-components/signpost/#positions)

lenriquez:
@GuessWhoSamFoo I think this solution will work for now

web/src/app/modules/shared/components/presentation/overflow-labels/overflow-labels.component.scss
```diff
@@ -43,9 +43,14 @@
...
   }
 
+  ::ng-deep &-content {
+    max-width: min-content;
+  }
+
   // To avoid horizontal scroll-bar inside of the signpost content body.
   ::ng-deep .signpost-wrap .signpost-content-body {
     padding: 1rem;
+    max-height: 12rem;
```
mklanjsek, line 53: Let's change this to `12.2rem` to make sure the last one is not cut off.

mklanjsek (commented):
Looks good for short term solution, until Clarity improves it on their side. I have only one small request about max height below.

mklanjsek (approved):

mklanjsek:
Please make sure to squash commits before merging.

mklanjsek:
Karma tests are still failing on Windows

Issue #1750, opened 2020-12-18
Remove rice box from build tools


GuessWhoSamFoo:
https://tip.golang.org/doc/go1.16#library-embed

Go 1.16 offers support for embedded files. This should simplify build steps and reduce friction for a development environment. As part of the dependency bump, we should:

- Clean rice-box.go (which is currently ignored by git)
- Update Github action workflows
- Remove reference from `tools.go`; update `go.mod`
- Re-bundle `dist/dash-frontend` currently handled in `web/web.go`

wwitzel3:
Blocked by the official release of Go 1.16.

If someone wants to try this against 1.16 beta and put it as a draft PR while we wait for the release, that would be great.

xtreme-vikram-yadav:
I picked this issue today

PR #2075, opened 2021-02-25
Fix build-electron, remove mkdir, not needed

wwitzel3:

Currently `go run build.go build-electron` is broken on Windows.
This removes the `webBuildElectron` function completely. After the recent changes to remove go-rice #1750 this is no longer neded.

### DIFF
--- a/build.go
+++ b/build.go
@@ -347,15 +347,6 @@ func webBuild() {
 }
 
 func webBuildElectron() {
-	dirCmd := newCmd("mkdir", nil, "-p", "dist/octant")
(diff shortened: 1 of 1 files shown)


GuessWhoSamFoo:
I think this is still needed because it removes bundled js  which would otherwise be packaged by electron-builder

wwitzel3:
You're right, but we don't need to run a `mkdir` first, the clean command works even if the directory does not exist.

mklanjsek (approved):

PR #2099, opened 2021-03-03
Add lookup for additional metadata from JS plugins

### LINKED ISSUE #2042: Plugins can enrich calls to custom object stores with arbitrary metadata
(co-written with @chinigo)

We're developing a custom `store.Store` implementation that retrieves Kubernetes API data through an authenticated proxy. The results of these API calls may vary based on the authenticated user making the query.  The auth flow is handled in the web browser and tokens are short-lived, so the frontend needs some way to propagate its credentials back to the store.

First, we'll call out the fact that our use case involves a plugin which makes use of the Javascript runtime, but will mention Go/GRPC plugins later.
​
[`octant.DashboardClient`](https://github.com/vmware-tanzu/plugin-library-for-octant/blob/7b9e6ad4dc1cd303d92ed3e0a906a7cbfc058f8a/plugin/octant.d.ts#L109-L158) could be modified such that each method has an extra metadata parameter, e.g.
​
```typescript
export interface DashboardClient {
  List(key: Key, metadata?: {[key: string]: string}) : any[];
  // ... other methods would add a similar optional metadata parameter
}
```
​
Then we could create a plugin that registers [a custom action](https://github.com/vmware-tanzu/octant/blob/v0.17.0/pkg/action/doc.go#L13) to keep track of the state of the client's authentication, and passes that data to its `dashboardClient` using the new optional metadata parameter:
​
```typescript
export default class TokenHandlingPlugin implements octant.Plugin {
  dashboardClient: octant.DashboardClient;
  httpClient: octant.HTTPClient;
  // ... plugin boilerplate ...
​
  tokens : {[clientID: string]: string};
​
  capabilities = {
    actionNames: ["action.mycompany.org/setAuthToken"],
  };
​
  actionHandler(request: octant.ActionRequest): octant.ActionResponse | void {
    if (request.actionName === "action.mycompany.org/setAuthToken") {
      this.tokens[request.clientID] = request.payload.token;
      return;
    }
    return;
  }
​
  contentHandler(request: octant.ContentRequest): octant.ContentResponse {
    let metadata = {};
    metadata["auth_token"] = this.tokens[request.clientID];
    let pods = this.dashboardClient.List(
      {apiVersion: 'v1', kind: 'Pod'},
      metadata // <----------------------------------------- NEW PARAMETER
    );
    return {content: {viewComponents: [], title: []}};
  }
}
```
​
Then the [corresponding function in the javascript plugin runtime](https://github.com/vmware-tanzu/octant/blob/fcd6330a1af86a0be57ed05e70c3104403192dc8/pkg/plugin/javascript/dashboard_list.go#L38-L67) could be modified to inject these values into the `context.Context` that gets passed to its associated `store.Store`, something like:
​
```go
func (d *DashboardList) Call(ctx context.Context, vm *goja.Runtime) func(c goja.FunctionCall) goja.Value {
	return func(c goja.FunctionCall) goja.Value {
		md := map[string]interface{}{}
		mdObj := c.Argument(1).ToObject(vm)
		_ = vm.ExportTo(mdObj, &md)
		for key, val := range md {
			ctx = context.WithValue(ctx, key, val)
		}
		// ... followed by the usual implementation
	}
}
```
​
Finally, our custom `store.Store` implementation could read the auth token from the context passed to its `List` method with `ctx.Val("auth_token")` and use it to authenticate with the aforementioned proxy.
​
In the case of GRPC plugins, a slightly different implementation would be required, where perhaps consumers of [`pkg/plugin/api.Client`](https://github.com/vmware-tanzu/octant/blob/master/pkg/plugin/api/client.go#L87-L107) could expect metadata values placed in the `context.Context` parameter to be serialized by the client, perhaps as GRPC headers, on the RPC backing each method. Then the [server's implementation](https://github.com/vmware-tanzu/octant/blob/fcd6330a1af86a0be57ed05e70c3104403192dc8/pkg/plugin/api/server.go#L87-L92) of those RPCs would be responsible for deserializing the metadata values and packing them into the `context.Context` passed to the associated `store.Store` in a similar manner to that described for the javascript runtime above.

Our use case is for authentication, but we can imagine other scenarios where it would be useful for a plugin to coordinate with a custom object store:
- Transport-layer concerns like correlation IDs, proxy configuration, or cache busting.
- As per discussion @wwitzel3, you could imagine an object store backed by a database, where the frontend client would need to manage cursor position in a result set.

**Environment:**

- Octant version (use `octant version`): 0.17.0
- Kubernetes version (use `kubectl version`): N/A
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): a [`golang`](https://hub.docker.com/_/golang) (i.e. Debian buster)-based container running on kubernetes. AWS is providing the infrastructure, but I'm unsure of whether it's an EKS cluster or maybe [Tanzu Kubernetes Grid](https://docs.vmware.com/en/VMware-Tanzu-Kubernetes-Grid/1.0/vmware-tanzu-kubernetes-grid-10/GUID-install-tkg-aws.html).

xtreme-vikram-yadav:
It allows octant to extract out additional contextual data sent by the JS plugins.

- Fixes #2042

**Special notes for your reviewer**: There will be a separate PR to add similar support for Go plugins.

Related PR: https://github.com/vmware-tanzu/plugin-library-for-octant/pull/8

### DIFF
--- a/pkg/plugin/javascript/dashboard_client.go
+++ b/pkg/plugin/javascript/dashboard_client.go
@@ -16,6 +16,15 @@ import (
 	"github.com/vmware-tanzu/octant/internal/octant"
 )
 
+// DashboardMetadataKey is a type used for metadata keys passed by plugins
+type DashboardMetadataKey string
+
+//DashboardMetadataSentinel is a type for CloneKey
+type DashboardMetadataSentinel string
+
+//CloneKey is a dummy key used to clone context.Context
+var CloneKey DashboardMetadataSentinel = ""
+
 // ModularDashboardClientFactory is a modular octant.DashboardClientFactory. It configures
 // itself based on functions passed when it is initialized.
 type ModularDashboardClientFactory struct {
--- a/pkg/plugin/javascript/dashboard_client.go
+++ b/pkg/plugin/javascript/dashboard_client.go
@@ -19,12 +19,6 @@ import (
 // DashboardMetadataKey is a type used for metadata keys passed by plugins
 type DashboardMetadataKey string
 
-//DashboardMetadataSentinel is a type for CloneKey
-type DashboardMetadataSentinel string
-
-//CloneKey is a dummy key used to clone context.Context
-var CloneKey DashboardMetadataSentinel = ""
-
 // ModularDashboardClientFactory is a modular octant.DashboardClientFactory. It configures
 // itself based on functions passed when it is initialized.
 type ModularDashboardClientFactory struct {
--- a/pkg/plugin/javascript/dashboard_client.go
+++ b/pkg/plugin/javascript/dashboard_client.go
@@ -75,3 +75,18 @@ func panicMessage(vm *goja.Runtime, err error, reason string) goja.Value {
 
 	return vm.ToValue(fmt.Sprintf("%s: %s", reason, err.Error()))
 }
+
+// extract out the key/values sent by the plugin to add context for object store requests.
+func setObjectStoreContext(ctx context.Context, jsObj goja.Value, vm *goja.Runtime) context.Context {
+	var metadata map[string]string
+	metadataObj := jsObj.ToObject(vm)
+
+	// This will not error as js plugins restrict this type
+	// and we handle the case of having undefined value
+	_ = vm.ExportTo(metadataObj, &metadata)
+	for k, val := range metadata {
+		ctx = context.WithValue(ctx, DashboardMetadataKey(k), val)
+	}
+
+	return ctx
+}
--- a/pkg/plugin/javascript/dashboard_delete.go
+++ b/pkg/plugin/javascript/dashboard_delete.go
@@ -49,15 +49,7 @@ func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goj
 
 		metadataArg := c.Argument(1)
 		if !goja.IsUndefined(metadataArg) {
-			var metadata map[string]string
-			metadataObj := metadataArg.ToObject(vm)
-
-			// This will not error as js plugins restrict this type
-			// and we handle both cases
-			_ = vm.ExportTo(metadataObj, &metadata)
-			for k, val := range metadata {
-				newCtx = context.WithValue(newCtx, DashboardMetadataKey(k), val)
-			}
+			newCtx = setObjectStoreContext(newCtx, metadataArg, vm)
 		}
 
 		if err := d.storage.ObjectStore().Delete(newCtx, key); err != nil {
--- a/pkg/plugin/javascript/dashboard_delete.go
+++ b/pkg/plugin/javascript/dashboard_delete.go
@@ -38,6 +38,7 @@ func (d *DashboardDelete) Name() string {
 // delete is unsuccessful, it will throw a javascript exception.
 func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goja.FunctionCall) goja.Value {
 	return func(c goja.FunctionCall) goja.Value {
+		var newCtx context.Context = context.WithValue(ctx, CloneKey, nil)
 		var key store.Key
 		obj := c.Argument(0).ToObject(vm)
 
@@ -49,12 +50,12 @@ func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goj
 			metadataObj := c.Argument(1).ToObject(vm)
 
 			_ = vm.ExportTo(metadataObj, &metadata)
-			for key, val := range metadata {
-				ctx = context.WithValue(ctx, key, val)
+			for k, val := range metadata {
+				newCtx = context.WithValue(newCtx, DashboardMetadataKey(k), val)
 			}
 		}
 
-		if err := d.storage.ObjectStore().Delete(ctx, key); err != nil {
+		if err := d.storage.ObjectStore().Delete(newCtx, key); err != nil {
 			panic(panicMessage(vm, err, ""))
 		}
 		return goja.Undefined()
--- a/pkg/plugin/javascript/dashboard_delete.go
+++ b/pkg/plugin/javascript/dashboard_delete.go
@@ -44,6 +44,16 @@ func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goj
 		// This will never error since &key is a pointer to a type.
 		_ = vm.ExportTo(obj, &key)
 
+		if len(c.Arguments) > 1 {
+			var metadata map[string]interface{}
+			metadataObj := c.Argument(1).ToObject(vm)
+
+			_ = vm.ExportTo(metadataObj, &metadata)
+			for key, val := range metadata {
+				ctx = context.WithValue(ctx, key, val)
+			}
+		}
+
 		if err := d.storage.ObjectStore().Delete(ctx, key); err != nil {
 			panic(panicMessage(vm, err, ""))
 		}
--- a/pkg/plugin/javascript/dashboard_delete.go
+++ b/pkg/plugin/javascript/dashboard_delete.go
@@ -38,17 +38,22 @@ func (d *DashboardDelete) Name() string {
 // delete is unsuccessful, it will throw a javascript exception.
 func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goja.FunctionCall) goja.Value {
 	return func(c goja.FunctionCall) goja.Value {
-		var newCtx context.Context = context.WithValue(ctx, CloneKey, nil)
+		newCtx, cancel := context.WithCancel(ctx)
+		defer cancel()
+
 		var key store.Key
 		obj := c.Argument(0).ToObject(vm)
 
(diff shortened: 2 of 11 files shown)


pkg/plugin/javascript/dashboard_delete.go
```diff
@@ -44,6 +44,16 @@ func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goj
 		// This will never error since &key is a pointer to a type.
 		_ = vm.ExportTo(obj, &key)
 
+		if len(c.Arguments) > 1 {
```
xtreme-vikram-yadav, line 47: This is needed to keep the metadata argument optional. I don't have strong opinions about this so let me know if anybody prefers this to always be passed with calls to the data store.

pkg/plugin/javascript/dashboard_delete.go
```diff
@@ -44,6 +44,16 @@ func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goj
...
+			var metadata map[string]interface{}
+			metadataObj := c.Argument(1).ToObject(vm)
+
+			_ = vm.ExportTo(metadataObj, &metadata)
+			for key, val := range metadata {
+				ctx = context.WithValue(ctx, key, val)
```
xtreme-vikram-yadav, line 53: Seems like the context passed in here has a long life and adding context here keep increasing the memory imprint of the context. While it doesn't impact the correctness of the functionality, its growth with every execution of this code is concerning to me.

xtreme-vikram-yadav: Example when i logged the context(I am passing datetime utc string from JS plugin in this example):

It started off with: 
```context : context.Background.WithCancel.WithValue(type log.key, val <not Stringer>).WithValue(type context.OctantContextKey, val <not Stringer>).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT)```

After a while the context grew:
```
context : context.Background.WithCancel.WithValue(type log.key, val <not Stringer>).WithValue(type context.OctantContextKey, val <not Stringer>).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT).WithValue(type string, val Wed, 03 Mar 2021 02:02:39 GMT)^C2021-03-02T21:03:19.268-0500```

chinigo: Is this because we're iterating over [*all* the keys of the metadata](https://github.com/vmware-tanzu/octant/blob/9491df7a283c07dbe434ec49ead9686a483f6460/pkg/plugin/javascript/dashboard_delete.go#L52), but only [deleting the *last* key](https://github.com/vmware-tanzu/octant/blob/9491df7a283c07dbe434ec49ead9686a483f6460/pkg/plugin/javascript/dashboard_delete.go#L57)?

It was my impression that [`context.WithValue` creates a copy](https://golang.org/pkg/context/#WithValue) rather than mutating the existing instance, so because, here, the copied `ctx` object goes out of scope at the end of this method, I would not have thought a memory leak was possible.

If it *is* a long-lived object, how does it behave when multiple calls specify different values for a given key?

pkg/plugin/javascript/dashboard_delete.go
chinigo: Javascript [doesn't care one way or the other](https://dmitripavlutin.com/javascript-function-parameters/), so either way seems fine to me.

chinigo (commented):
A handful of questions but no objections.

pkg/plugin/javascript/dashboard_delete.go
xtreme-vikram-yadav: I was thinking of a scenario where octant users decide to use the older plugins with new octant release. This will not introduce any breaking change in that case

pkg/plugin/javascript/dashboard_delete.go
xtreme-vikram-yadav: I should be using a different variable name than `key` here to avoid confusion

xtreme-vikram-yadav: It's because i'm reusing the variable for the new context. I'm experimenting with a new variable for each function call to avoid this.

xtreme-vikram-yadav:
I'm syncing with @chinigo and @dimitrybesson to address some of the concerns.

pkg/plugin/javascript/dashboard_delete.go
```diff
@@ -38,13 +38,24 @@ func (d *DashboardDelete) Name() string {
 // delete is unsuccessful, it will throw a javascript exception.
 func (d *DashboardDelete) Call(ctx context.Context, vm *goja.Runtime) func(c goja.FunctionCall) goja.Value {
 	return func(c goja.FunctionCall) goja.Value {
+		var newCtx context.Context = context.WithValue(ctx, CloneKey, nil)
```
wwitzel3, line 41: This is confusing at a glance, I'd rather see

    // or optionally ctx, _ :=
    ctx, cancel := context.WithCancel(ctx)
    defer cancel()

pkg/plugin/javascript/dashboard_client.go
```diff
@@ -16,6 +16,15 @@ import (
...
 )
 
+// DashboardMetadataKey is a type used for metadata keys passed by plugins
+type DashboardMetadataKey string
+
+//DashboardMetadataSentinel is a type for CloneKey
+type DashboardMetadataSentinel string
+
+//CloneKey is a dummy key used to clone context.Context
+var CloneKey DashboardMetadataSentinel = ""
```
wwitzel3, line 26: These can go away if we use `WithCancel` to make a copy. Also it avoids some potential confusion when a plugin author tries to send it data and use empty string (which is a valid key) as their key.

pkg/plugin/javascript/dashboard_list.go
```diff
@@ -52,7 +53,17 @@ func (d *DashboardList) Call(ctx context.Context, vm *goja.Runtime) func(c goja.
 			panicMessage(vm, fmt.Errorf("key is invalid: %w", err), "")
 		}
 
-		u, _, err := d.storage.ObjectStore().List(ctx, key)
+		if len(c.Arguments) > 1 {
+			var metadata map[string]interface{}
+			metadataObj := c.Argument(1).ToObject(vm)
+
+			_ = vm.ExportTo(metadataObj, &metadata)
```
wwitzel3, line 60: always good to explain why the code is throwing away the error

wwitzel3 (changes requested):
This is looking great, I'd like to see a couple minor updates and then I think this is ready.

wwitzel3 (approved):

wwitzel3:
This is an interesting example of a similar idea, https://github.com/grpc/grpc-go/blob/v1.36.0/metadata/metadata.go#L154 .. I think I would like to see the Key/Value extraction and copying done as a function that we can unit test.

wwitzel3 (changes requested):
Sorry @xtreme-vikram-yadav I changed my mind, I would like to see the Key/Value extraction and copy process as a stand-alone function.

I don't think we need to use the go-grpc implementation, but I don't like that pattern of creating a unit testable single purpose function, instead of copying and pasting this in to each client method.

wwitzel3 (approved):
LGTM

PR #2102, opened 2021-03-05
Switch to jsoniter

wwitzel3:
@GuessWhoSamFoo made a great point in the last community meeting that we needed to develop a strategy to address CPU issues we are seeing in Octant. I already had this branch in flight and in my testing I'm seeing good enough improvements generally to make the external dependency worth it.

## Note
This is not an attempt to fix the high CPU spikes that have been reported on macOS. This was more a general exploration that I didn't want to die on the vine.

I'm also open to not landing this, but the way I see it, https://github.com/json-iterator/go is a standard library compatible well maintained JSON library that gives us speed improvements with zero effort.

These tests were captured against the same namespace, cluster, and conditions for each binary and the same actions were performed.

### Before *encoding/json* (grabbing the middle avg)
image

### After *jsoniter* (grabbing the middle avg)
image

Signed-off-by: Wayne Witzel III <wayne@riotousliving.com>

PR #2105, opened 2021-03-05
Memory caching for CRD navigation entries

mklanjsek:
During performance testing, I noticed that we are rebuilding a list of CRD navigation entries twice per second, for cluster scope and current namespace. That is very costly operation taking ~500ms for larger CRD sets. To address the issue, I added a memory cache holding CRD navigation entries. Cache is invalidated every time new CRD appears or goes away.

Performance improvement is considerable but it depends on number of CRDs in cluster/namespace - I am seeing 15-20% improvement on larger sets on my mac.


wwitzel3:
Good start, this will need to be tied in to the Add/Update/Delete events of crd_watcher.go so that it removes old CRDS and adds new CRDs as they are applied. This will only cache the CRDs that are installed at the time you startup Octant.

mklanjsek:
@wwitzel3 I already handle Add/Delete in `crd_watcher`, invalidating the cache there so it's recreated during next call.

wwitzel3:
> @wwitzel3 I already handle Add/Delete in `crd_watcher`, invalidating the cache there so it's recreated during next call.

I see it now, probably was just GitHub not showing me the whole diff.

wwitzel3:
Current:
image

This change alone:
image

This change combined with the #2102 branch
image

wwitzel3 (changes requested):
This introduces a data race since many go routines can be calling `CRDEntries`, compiling with the `-race` flag reports a race with the sort.Slice in pkg/navigation/navigation.go 88-90

mklanjsek:
Good catch Wayne, missed the race condition.  I eliminated the redundant sort on cached object and as a result the race condition is gone now.

Discouraged by negligible performance gain in your tests, makes me wonder if it's worth it...

wwitzel3:
Negligible? Avg. 50% improvement once cached, that's a big improvement.

mklanjsek:
Ooops my bad, didn't read that well 🤦  Definitely not negligible!

wwitzel3:
I want to think about this a bit more, the idea is solid and effective. I'm not sure I like the current implementation of just slapping a global in to navigation and calling it from the watcher callbacks and accessing it in `CRDEntires` and `CustomResourceDefinitions`.

Also there will need to be at least a simple assertion test: non-cache call, cache call, clear cache, non-cache call, cache call.

wwitzel3:
@mklanjsek and I talked about this, we will implement a cache layer and then overview and custom overview will maintain their own cache of the `CRDEntries` call. This is because the modules actually care about different results and the CRDEntires function is bi-modal depending of if you want namespace scoped CRDs or not.

mklanjsek:
Waiting until #2144 is merged before finishing this.

mklanjsek:
@wwitzel3 PTAL

wwitzel3:
Please rebase, there is a failing test when merged with master because we no longer use `RegisterOnUpdate` methods.

mklanjsek:
Rebased and fixed test failure.

wwitzel3 (approved):

Issue #2013, opened 2021-02-17
Even when doing nothing, Octant burns cpu cycles


scothis:
**What steps did you take and what happened:**

Running the election MacOS app, close the octant UI and observe the resource utilization for octant. The background process is consuming between 120-160% of a cpu core. This is pure backend load even when there is no client connected.

<img width="1072" alt="Screen Shot 2021-02-16 at 8 13 50 PM" src="https://user-images.githubusercontent.com/302992/108142726-5183e080-7094-11eb-9257-4c8968f3e9da.png">

**What did you expect to happen:**

I'd expect the backend to use some CPU cycles just to keep the caches alive for when the front end connects. Maybe 5-10% of a CPU core is acceptable.

**Environment:**

- Octant version (use `octant version`): 0.17.0
- Kubernetes version (use `kubectl version`): 1.17
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): macOS 11.2.1 (Rossetta)

wwitzel3:
I'm going to close this resolved, but please re-open if you are still seeing Octant use a large amount of CPU even when minimized to the system tray.

PR #2127, opened 2021-03-09
Add memstat flag

GuessWhoSamFoo:
Allows users to benchmark memory usage without needing additional setup

- related to #2013 

Command can be invoked by:
`octant --disable-open-browser --memstats stats.txt --meminterval 500ms`

then observe via `tail -f stats.txt`


wwitzel3 (approved):
Thanks, this is useful. In the future lets think about how we can start to expose these knobs easily through our preferences. We have a developer section that can have flags like this.

PR #2128, opened 2021-03-09
Fix for security vulnerabilities

mklanjsek:
Addressed a couple of reported security vulnerabilities:
- Updated `marked` library to version 2.0.1 to address the  [marked security vulnerability]( https://github.com/vmware-tanzu/octant/security/dependabot/web/package.json/marked/open)
- Bumped the `ini` dependency to address the [ini prototype pollution vulnerability](https://www.npmjs.com/advisories/1589)

Note that the build process now has pre-install phase that performs dependency resolutions - that is the only way to update the nested dependencies with `npm`

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -16621,9 +16621,9 @@
       "dev": true
     },
     "ini": {
-      "version": "1.3.5",
-      "resolved": "https://registry.npmjs.org/ini/-/ini-1.3.5.tgz",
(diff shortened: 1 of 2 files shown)


GuessWhoSamFoo:
Out of curiosity, why is resolutions needed here rather than bumping marked to 2.x?

mklanjsek:
Resolutions section is needed to override the secondary dependency that `@compodoc/compodoc` has on `marked`.  Without that, running `npm install` would always bring it back to version 0.7.0.

wwitzel3 (approved):

PR #2137, opened 2021-03-10
Fixed problem with jsoniter sorting

### LINKED ISSUE #2133: Labels causing layout shift and performance degradation
When kubernetes resources containing multiple labels are present, labels are sometimes updated in Octant UI without any changes in underlining label data. It's most likely happening because internal ordering of labels changes, causing different labels to move to the start of the list. This is causing layout shifts and affecting both user experience and screen update performance. It appears to be happening periodically after every ~10 seconds.

Here is an example of layout shift happening around 2:30 in recording:
https://user-images.githubusercontent.com/34245594/110654112-13ea1300-817b-11eb-80c1-3fcadf7e12d2.mov

As illustrated in performance dump, there are two layout shifts happening in this scenario, first is 41.5ms and second one 121.7ms long:
Screen Shot 2021-03-10 at 8 21 19 AM

The [unnecessary layout shifts](https://web.dev/cls/?utm_source=devtools) are degrading user experience and we should try to eliminate them.

mklanjsek:
- Refactored the `jsoniter` initialization code inside a single entry point in global utility package,
- Modified the `jsoniter` configuration to use combination of `ConfigFastest ` and `SortMapKeys ` flags,
- Verified that addressed both label and resource viewer issue from #2133,
- Added unit test to make sure json marshaling is properly sorting fields, 

Fixes #2133

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -13,8 +13,10 @@ import (
 	"strings"
 	"time"
 
+	"github.com/vmware-tanzu/octant/internal/util/json"
+
 	"github.com/gorilla/mux"
-	jsoniter "github.com/json-iterator/go"
+
 	"github.com/spf13/viper"
 
 	"github.com/vmware-tanzu/octant/internal/config"
@@ -24,8 +26,6 @@ import (
 )
 
 //go:generate mockgen -destination=./fake/mock_service.go -package=fake github.com/vmware-tanzu/octant/internal/api Service
-
-var json = jsoniter.ConfigFastest
 
 const (
 	// ListenerAddrKey is the environment variable for the Octant listener address.
--- a/internal/api/api_test.go
+++ b/internal/api/api_test.go
@@ -16,7 +16,6 @@ import (
 	"testing"
 
 	"github.com/golang/mock/gomock"
-	jsoniter "github.com/json-iterator/go"
 	"github.com/pkg/errors"
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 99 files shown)


wwitzel3 (changes requested):
This is not how we want to do this.

If we want to centralize the configuration, that's fine, but this isn't the way.

wwitzel3:
If we want to do this, define our own internal `json` package, like this.

```
diff --git a/internal/json/json.go b/internal/json/json.go
new file mode 100644
index 00000000..2c3c26ca
--- /dev/null
+++ b/internal/json/json.go
@@ -0,0 +1,15 @@
+package json
+
+import jsoniter "github.com/json-iterator/go"
+
+var jsonConfig = jsoniter.Config{
+       EscapeHTML:                    false,
+       MarshalFloatWith6Digits:       true,
+       ObjectFieldMustBeSimpleString: true,
+       SortMapKeys:                   true,
+}.Froze()
+
+var NewEncoder = jsonConfig.NewEncoder
+var Unmarshal = jsonConfig.Unmarshal
+var Marshal = jsonConfig.Marshal
+var MarshalIndent = jsonConfig.MarshalIndent
```

mklanjsek:
Thanks Wayne. I generally try to avoid shadowing packages, but on the second look I think you are right, that's better way of doing it.

mklanjsek:
@wwitzel3 PTAL

wwitzel3 (approved):
Perfect, thank you @mklanjsek

PR #2165, opened 2021-03-16
Allow plugins to add multiple tabs

### LINKED ISSUE #1440: Allow plugins to contribute multiple tab responses

Octant is great and the extensions API allows me to add a single new tabs for the specified Kubernetes object. For example, in [Starboard plugin](https://github.com/aquasecurity/starboard-octant-plugin) we add the **Vulnerabilities**, which contains vulnerability reports for the given workload containers.

plugin_multi_tabs

Starboard also checks deployment descriptors for insecure configurations and we'd like to add another tab, next to the Vulnerabilities tab, which contains config audit report and is called **Configuration Audit**. 

- We could always add a single generic tab called maybe Starboard, which contains tabbed pane inside, but this hides the intention of the tab or UI component contributed by the plugin.
- Support for multiple top level tabs will allow users to spot new items they can click and experiment with.
- It would be nice to have it in Go and Typescript APIs as we haven't migrated to Typescript yet

**Environment:**

GuessWhoSamFoo:
This is a non-breaking API change that allows plugin authors to add multiple tabs from a single plugin. Plugins will still be displayed in lexical order as they are sorted with tabs potentially from other plugins.

```
options := []service.PluginOption{
	service.WithTabPrinter(handleTab, handleTab2, handleTab3),
}
```

- Fixes #1440 


pkg/plugin/javascript.go
```diff
@@ -390,14 +390,16 @@ func (t *jsPlugin) PrintTab(ctx context.Context, object runtime.Object) (TabResp
...
 		}
 		realFlexLayout := *realComponent.(*component.FlexLayout)
 		cTab.Contents.Config = realFlexLayout.Config
 	}
 
-	return TabResponse{
-		Tab: cTab,
+	return []TabResponse{
+		{
+			Tab: cTab,
```
xtreme-vikram-yadav, line 401: So for now JS plugins can still generate a single tab? I'm guessing there will be updates here after corresponding changes are made pushed https://github.com/vmware-tanzu/plugin-library-for-octant?

wwitzel3: Good call out, we'll want to update https://github.com/vmware-tanzu/plugin-library-for-octant/blob/main/plugin/helpers.ts#L83 and https://github.com/vmware-tanzu/plugin-library-for-octant/blob/main/plugin/octant.d.ts#L62 after this lands.

wwitzel3 (approved):
This looks good to me, please create an issue in the plugin library repo to ensure we update the libraries there for this new behavior as part of 0.18.

As mentioned in a comment, we'll want to update https://github.com/vmware-tanzu/plugin-library-for-octant/blob/main/plugin/helpers.ts#L83 and https://github.com/vmware-tanzu/plugin-library-for-octant/blob/main/plugin/octant.d.ts#L62 after this lands.

GuessWhoSamFoo:
Merging so npm libraries can cut an update if desired.

wwitzel3:
Bad review from me previously, I completely missed that the `TabResponse` object in the jsPlugin.PrintTabs method was hard coded to a single tab. So even with the typescript updates, the JS plugins won't be able to generate multiple tabs.

Sorry about that, I'll open for that and see about getting that implemented along side the TS libraries.

https://github.com/vmware-tanzu/octant/blob/master/pkg/plugin/javascript.go#L399

PR #2179, opened 2021-03-18
Add expandable rows to datagrids

### LINKED ISSUE #2019: Add row detail from Clarity
As a plugin author,
Given that I create a table
I would like to let the user expand and collapse each row
So that they can see row details

Screen Shot 2021-02-17 at 2 22 04 PM

GuessWhoSamFoo:

Peek 2021-03-18 16-19

Allows rows in datagrids to have expandable rows which contains another view component.

- Fixes #2019 

### DIFF
--- a/web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.ts
+++ b/web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.ts
@@ -12,12 +12,14 @@ import {
 } from '@angular/core';
 import {
   Confirmation,
+  ExpandableRowDetailView,
   GridAction,
   GridActionsView,
   TableFilters,
   TableRow,
   TableRowWithMetadata,
   TableView,
+  View,
 } from 'src/app/modules/shared/models/content';
 import trackByIndex from 'src/app/util/trackBy/trackByIndex';
 import trackByIdentity from 'src/app/util/trackBy/trackByIdentity';
@@ -54,6 +56,7 @@ export class DatagridComponent
   buttonGroup?: ButtonGroupView;
   isModalOpen = false;
   defaultPageSize: number;
+  detail = { name: 'test' };
 
   actionDialogOptions: ActionDialogOptions = undefined;
 
@@ -111,9 +114,17 @@ export class DatagridComponent
 
     return rows.map(row => {
       let actions: GridAction[] = [];
+      let detailView: View;
+      let replace: boolean;
 
       if (row.hasOwnProperty('_action')) {
         actions = (row._action as GridActionsView).config.actions;
+      }
+
+      if (row.hasOwnProperty('_expand')) {
+        const erv = (row._expand as ExpandableRowDetailView).config;
+        detailView = erv.body;
+        replace = erv.replace;
       }
 
       const isDeleted = !!row._isDeleted;
@@ -122,6 +133,8 @@ export class DatagridComponent
         data: row,
         actions,
         isDeleted,
+        detailView,
+        replace,
       };
     });
   }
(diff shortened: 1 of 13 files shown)


wwitzel3:
Let's add a storybook example under table for this.

GuessWhoSamFoo:
A clarity table listed under a storybook table does not have expandable rows. Octant calls datagraids as tables which can be confusing which is why there isn't a datagrid entry in storybook either. We will have to decide either to rename or explain the table/datagrid semantics in the table storybook entry.

wwitzel3:
> A clarity table listed under a storybook table does not have expandable rows. Octant calls datagraids as tables which can be confusing which is why there isn't a datagrid entry in storybook either. We will have to decide either to rename or explain the table/datagrid semantics in the table storybook entry.

Not sure I follow? .. There are two stories under Storybook -> Table one uses DatagridComponent and the other TableComponent , we can't add the expandable row to the one using the DatagridComponent?

GuessWhoSamFoo:
Nevermind, confused table and datagrid with incorrect copypasta

web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.ts
```diff
@@ -54,6 +56,7 @@ export class DatagridComponent
   buttonGroup?: ButtonGroupView;
   isModalOpen = false;
   defaultPageSize: number;
+  detail = { name: 'test' };
```
mklanjsek, line 59: `test` is not a good default value

web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.ts
```diff
@@ -111,17 +114,27 @@ export class DatagridComponent
...
+        replace = erv.replace;
+      }
+
       const isDeleted = !!row._isDeleted;
 
       return {
         data: row,
         actions,
         isDeleted,
+        detailView,
```
mklanjsek, line 136: `TableRowWithMetadata` needs to be updated with new fields

web/src/stories/table.stories.mdx
```diff
@@ -11,6 +11,18 @@ table.Add(component.TableRow{
 table.Sort("A")
 table.Reverse()`}}
 
+export const tableExpandableRowDocs= {source: {code: `cols := component.NewTableCols("A", "B")
```
mklanjsek, line 14: This story needs more work: 
- please add more details about this feature to docs section, 
- the size of each section can be uniformly reduced,
- section headers need to be added,
- there should be only one Props section at the bottom,

See for example the docs page for Dropdown component

mklanjsek (changes requested):
Looking good, just a few small things.

web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.ts
GuessWhoSamFoo: This line was a leftover from development and is not needed.

web/src/stories/table.stories.mdx
GuessWhoSamFoo: There has to be two props since `Table` can mean two different components. The remaining sections are updated to match the Dropdown section as much as possible.

web/src/app/modules/shared/components/presentation/datagrid/datagrid.component.ts
GuessWhoSamFoo: What new fields? An expandable row exists as a `"_expand":View` pair on TableRow.

mklanjsek: Sorry, that was not clear... if `getRowsWithMetadata()` and `rowsWithMetadata` are declared as `TableRowWithMetadata[]`, then `TableRowWithMetadata` now needs `detailView` and `replace` fields added.

web/src/stories/table.stories.mdx
mklanjsek: Looks much better now, thanks! Very small detail: the docs section would look better if height of 1st table is ~250px and 2nd ~350.

mklanjsek (approved):

PR #2186, opened 2021-03-19
Add Icon Component

### LINKED ISSUE #2170: Create IconComponent
At the moment there is no way to create an Icon, It will be nice to have be able to use the clarity icons https://clarity.design/foundation/icons/api/

Create IconComponent

Necessary for issue #2018

**Environment:**

- Kubernetes version (use `kubectl version`):
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):

lenriquez:

- Fixes #2170 

We are using this API 
https://v4.clarity.design/icons/how-to-use
and not this 
https://clarity.design/foundation/icons/api/
**Release note**:
```
release-note
```

### DIFF
--- a/web/src/app/modules/shared/components/presentation/icon/icon.component.ts
+++ b/web/src/app/modules/shared/components/presentation/icon/icon.component.ts
@@ -1,0 +1,115 @@
+import {
+  ChangeDetectionStrategy,
+  ChangeDetectorRef,
+  Component,
+} from '@angular/core';
+import { AbstractViewComponent } from '../../abstract-view/abstract-view.component';
+import { IconView } from '../../../models/content';
+
+@Component({
+  selector: 'app-view-icon',
+  templateUrl: './icon.component.html',
+  styleUrls: ['./icon.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
+})
+export class IconComponent extends AbstractViewComponent<IconView> {
+  shape: string;
+  flip: string;
+  size: string;
+
+  styleAttr = '';
+  classAttr = '';
+
+  sizeHash = { sm: '16', md: '24', lg: '36"', xl: '48', xxl: '64' };
+  directionHash = {
+    up: 'transform: rotate(0deg);',
+    right: 'transform: rotate(90deg);',
+    down: 'transform: rotate(180deg);',
+    left: 'transform: rotate(270deg);',
+  };
+  statusHash = {
+    info: 'is-info ',
+    success: 'is-success ',
+    warning: 'is-warning ',
+    danger: 'is-error ',
+  };
+  badge = {
+    info: 'has-badge--info ',
+    success: 'has-badge--success ',
+    danger: 'has-badge ',
+    'warning-triangle': 'has-alert ',
+  };
+
+  constructor(private cdr: ChangeDetectorRef) {
+    super();
+  }
+
+  // This will work for version 5 of clarity but we are using
+  // version 4 so function are been added to map the differences
+  // dir needs to be change to direction on the template
+  protected update(): void {
+    const view = this.v;
+
+    // reset values to be re-calculated
+    this.styleAttr = '';
+    this.classAttr = '';
+
+    this.shape = view.config.shape;
+    this.flip = view.config.flip;
+    this.mapSize(view.config.size);
+    this.mapDirection(view.config.direction);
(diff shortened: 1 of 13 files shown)


lenriquez:
This should be ready for review

wwitzel3 (changes requested):
This isn't quite right. The badge isn't rendering properly.

```
{
  "metadata": {
    "type": "icon"
  },
  "config": {
    "shape": "bell",
    "size": "22",
    "solid": true,
    "status": "none",
    "inverse": false,
    "badge": "danger"
  }
}
```

Gives me: 
image

I'd expect, a red danger badge but in this case, whole icon is red.
image

Also I think `status` is too specific. Lets go with color and we can make some static types for our Success, Warning, and Failure status colors.

wwitzel3:
Another question, is this component useful as a stand-alone component like this? Or do we want to extend the other components to support an optional Icon as well?

For example if I want to write something like `My Link <Icon>` .. not sure how we do that with this currently?

I know this PR is just implementing the raw component, but I want to make sure we think about this as we build other things that use the icon component.

lenriquez:
@wwitzel3 Thanks for finding the bug! I will fix this tomorrow
I was thinking not on extending but rather on built a component that put the two component together. Something like "decorateComponent" and add the link and Icon to it `decorateComponent(Link, Icon)` with some configuration to show where the icon should be (left or right) and maybe the space should use. I am not sure about the name of the new component yet

GuessWhoSamFoo:
Should we hold this until we can upgrade to Clarity 5?

web/src/app/modules/shared/components/presentation/icon/icon.component.ts
```diff
@@ -0,0 +1,104 @@
...
+export class IconComponent extends AbstractViewComponent<IconView> {
+  shape: string;
+  flip: string;
+  size: string;
+
+  styleAttr = '';
+  classAttr = '';
+
+  sizeHash = { sm: '16', md: '24', lg: '36"', xl: '48', xxl: '64' };
+  directionHash = {
```
wwitzel3, line 24: After this merges, please create an issue that references all of these locations we can remove these mappings once we migrate from `clr-icon` to `cds-icon`

web/src/app/modules/shared/models/content.ts
```diff
@@ -585,3 +585,16 @@ export interface StepperView extends View {
...
 }
+
+export interface IconView extends View {
+  config: {
+    shape: string;
+    size: string;
+    direction: string;
+    flip: string;
+    solid: boolean;
+    status: string;
```
wwitzel3, line 596: I want to change this field to `color` and then create some static helpers called `SuccessColor` , `WarningColor` , `DangerColor` .. I think limiting this to just for colors will be a problem.

wwitzel3 (changes requested):

web/src/app/modules/shared/components/presentation/icon/icon.component.ts
lenriquez: Sure!

web/src/app/modules/shared/models/content.ts
lenriquez: by static helper do you mean change the mapper for a switch case? I mean 
rather than this
```
 statusHash = {
    info: 'is-info ',
    success: 'is-success ',
    warning: 'is-warning ',
    danger: 'is-error ',
  };
   mapStatus(status: string) {
    if (!status || status === '') {
      return;
    }
    this.classAttr += this.statusHash[status] || '';
  }
```
use this

```

  static SuccessColor = 'is-success';

  mapColor() {
    switch () {
      case: 'success'
        return this.SuccessColor;
    }
    .
    .
     .
  }

```

lenriquez: BTW Clarity 5 does a property call colors
https://clarity.design/foundation/icons/api/
image

lenriquez: Done!
image

wwitzel3: If both `color` and `status` are set, ignore `color` and use `status` .. other than that, this is great!

lenriquez: ohh I misunderstood that, It should work now
image

wwitzel3 (approved):

Issue #2184, opened 2021-03-19
Allow styling configuration for buttons 


GuessWhoSamFoo:
**Describe the problem/challenge you have**
Buttons in Octant are currently styled based on CSS rules from the frontend.

**Describe the solution you'd like**
For a given button, add the ability to:

- choose between solid, outlined, flat (no outline) styles
- disable a button
- create a [block button](https://clarity.design/angular-components/button/#block)
- show a button as loading
- allow setting colors to red, green, blue (error, success, info)

ftovaro:
Since @GuessWhoSamFoo did a big progress of this issue on this PR #2208 , I can take it from there and finish it

PR #2208, opened 2021-03-22
Add button styling and href link options

### LINKED ISSUE #2183: Buttons should be able to link
Currently buttons are only used to perform actions.

Ability to create a button which is really a `Link` component in disguise. Alternatively, redirecting to a certain path based on an action seems like a common use case.

GuessWhoSamFoo:
This PR create two methods for generating a link:

1. It allows plugin authors to wrap a component with `<a href=""><component></component></a>`.
2. It allows creating a link via `component.NewButtonLink`

It also creates a set of configurable styles based on clarity configurations. Due to previous usage of buttons, a default Octant button (not used in a form) is Clarity's `btn-small btn-outline`. The size is converted such that an Octant large button is a normal clarity button.

- Fixes #2183 and #2184 

### DIFF
--- a/pkg/view/component/button.go
+++ b/pkg/view/component/button.go
@@ -8,6 +8,39 @@ import (
 	"github.com/vmware-tanzu/octant/pkg/action"
 )
 
+const (
+	// ButtonStatusSuccess is a green button
+	ButtonStatusSuccess ButtonStatus = "success"
+	// ButtonStatusInfo is the default blue color
+	ButtonStatusInfo ButtonStatus = "info"
+	// ButtonStatusDanger is a red button
+	ButtonStatusDanger ButtonStatus = "danger"
+	// ButtonStatusDisabled is a disabled button
+	ButtonStatusDisabled ButtonStatus = "disabled"
+)
+
+type ButtonStatus string
+
+const (
+	// ButtonSizeBlock is a full width button
+	ButtonSizeBlock ButtonSize = "block"
+	// ButtonSizeSmall is a small button
+	ButtonSizeLarge ButtonSize = "lg"
+)
+
+type ButtonSize string
+
+const (
+	// ButtonStyleOutline is a transparent button with a colored border
+	ButtonStyleOutline ButtonStyle = "outline"
+	// ButtonStyleSolid is a button with solid color
+	ButtonStyleSolid ButtonStyle = "solid"
+	// ButtonStyleFlat is a button with no background color or outline
+	ButtonStyleFlat ButtonStyle = "link"
+)
+
+type ButtonStyle string
+
 // Confirmation is configuration for a confirmation dialog.
 type Confirmation struct {
 	Title string `json:"title"`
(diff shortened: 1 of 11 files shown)


pkg/view/component/button.go
```diff
@@ -88,6 +161,20 @@ func NewButton(name string, payload action.Payload, options ...ButtonOption) But
 	return button
 }
 
+// NewButtonLink creates a button component that navigates to a link
+func NewButtonLink(name, ref string, options ...ButtonOption) Button {
```
wwitzel3, line 165: Not sure if I like the `NewButtonLink` .. it kind of confuses the issue of "this is a button with an action" and "this is a link that is a button". I'm willing to have my mind changed though ...

pkg/view/component/button.go
```diff
@@ -72,6 +141,10 @@ type Button struct {
 	Payload      action.Payload `json:"payload"`
 	Confirmation *Confirmation  `json:"confirmation,omitempty"`
 	Modal        Component      `json:"modal,omitempty"`
+	Ref          string         `json:"ref,omitempty"`
```
wwitzel3, line 144: I don't think we should add `Ref` to buttons.

web/src/app/modules/shared/components/presentation/button-group/button-group.component.ts
```diff
@@ -67,6 +83,52 @@ export class ButtonGroupComponent extends AbstractViewComponent<ButtonGroupView>
       this.buttonLoad.emit(true);
       this.doAction(payload);
     }
+    if (ref) {
+      if (isUrlExternal(ref)) {
```
wwitzel3, line 87: I think all of this special case logic can go away if we prefer using `LinkComponent` to wrap a button.

pkg/view/component/link.go
```diff
@@ -19,8 +19,9 @@ var _ Component = &Link{}
 
 // LinkConfig is the contents of Link
 type LinkConfig struct {
-	Text string `json:"value"`
-	Ref  string `json:"ref"`
+	Text    string    `json:"value"`
+	Ref     string    `json:"ref"`
+	Content Component `json:"content,omitempty"`
```
wwitzel3, line 24: I'd like to see a comment and/or documentation around priority: if you provide both Content and Text, you're Content your Text will get swallowed up.

web/src/stories/button.group.stories.mdx
```diff
@@ -12,19 +12,28 @@ buttonGroup.AddButton(
       )))
 `}}
```
wwitzel3, line 13: I think I'd prefer to see stories under `LinkComponent` show how to add a button as a link.

wwitzel3 (changes requested):
All in all, this looks really good. The I like the button types for styling and the `With` approach for easily providing configuration options to the helper.

web/src/app/modules/shared/components/presentation/button-group/button-group.component.ts
GuessWhoSamFoo: Oddly enough, buttons are the exception to this because they are not components. Using the method in this PR, the entire button group would be wrapped as a link rather than individual buttons.

wwitzel3: AFAICT there is no good reason why Button isn't also a component type. It is just legacy and because "we never used them that way" .. but this is a good case for making button be stand-alone component that isn't tied to a group.

wwitzel3: Example: https://gist.github.com/wwitzel3/82fb4d1e0b6dad1fdb9bb204ff03214c

At least at first pass, it seem fine to make it a proper component. The current usage of button groups for components doesn't need to change, but it then allows folks to more generically use this new button component.

GuessWhoSamFoo:
Closed and to be resumed by @ftovaro

PR #2177, opened 2021-03-18
Resource Viewer: updated visual styles

mklanjsek:
Updated resource visual stiles based on #2010 and added set of storybook stories for easier testing of this and future iterations.

Note: also fixes the annoying storybook theme issues. 

Screen Shot 2021-03-18 at 8 03 01 AM

Fixes #2066

Signed-off-by: Milan Klanjsek mklanjsek@pivotal.io

PR #2212, opened 2021-03-24
Update component dynamically if view type changes

mklanjsek:
The recent Resource Viewer [change](https://github.com/vmware-tanzu/octant/pull/2177) uncovered problems with updating the object status component. After Resource Viewer selection has been changed, the status view was not updated when component of different type had to be used, so for example we were missing heptagon indicators in some cases. This change is fixing that by checking the view type and forcing the update if new view has different type.  


mklanjsek:
@GuessWhoSamFoo I believe this should fix the problems you were seeing with disappearing heptagons

GuessWhoSamFoo (approved):

PR #2221, opened 2021-03-25
Add Signpost

### LINKED ISSUE #2018: Implement signpost component from Clarity
As a plugin author
I would like to create fly out text
So that a user can see error messages when hovering

Screen Shot 2021-02-17 at 2 22 04 PM

lenriquez:

We need a way of adding a icon and sign post to any component

- Fixes #2018

This is how it looks like 
image

- I am not really sure if we need the mark down flag
- I think we should wait for clarity 5 PR to merge to test again

**Release note**:
```
release-note
```

### DIFF
--- a/pkg/view/component/signpost.go
+++ b/pkg/view/component/signpost.go
@@ -1,0 +1,95 @@
+/*
+Copyright (c) 2021 the Octant contributors. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package component
+
+import (
+	"github.com/vmware-tanzu/octant/internal/util/json"
+)
+
+// Signpost is a component for a signpost
+// +octant:component
+type Signpost struct {
+	Base
+	Config SignpostConfig `json:"config"`
+}
+
+// SignpostConfig is the contents of a signpost
+type SignpostConfig struct {
+	// Trigger is the component that will trigger the signpost.
+	Trigger Component `json:"trigger"`
+	// Message is the text that will be displayed.
+	Message string `json:"message"`
+	// Position of Signpost
+	Position Position `json:"position"`
+}
+
+type Position string
+
+const (
+	PositionTopLeft      Position = "top-left"
+	PositionTopMiddle    Position = "top-middle"
+	PositionTopRight     Position = "top-right"
+	PositionRightTop     Position = "right-top"
+	PositionRightMiddle  Position = "right-middle"
+	PositionRightBottom  Position = "right-bottom"
+	PositionBottomRight  Position = "bottom-right"
+	PositionBottomMiddle Position = "bottom-middle"
+	PositionBottomLeft   Position = "bottom-left"
+	PositionLeftBottom   Position = "left-bottom"
+	PositionLeftMiddle   Position = "left-middle"
+	PositionLeftTop      Position = "left-top"
+)
+
+func NewSignpost(c Component, t Component, m string) *Signpost {
+	so := &Signpost{
+		Base: newBase(TypeSignpost, nil),
+		Config: SignpostConfig{
+			Trigger: t,
+			Message: m,
+		},
+	}
+
+	return so
+}
+
+// SetStatus sets the status of the text component.
+func (t *Signpost) SetPosition(position Position) {
+	t.Config.Position = position
+}
+
+type signpostMarshal Signpost
+
+// MarshalJSON implements json.Marshaller
+func (t *Signpost) MarshalJSON() ([]byte, error) {
+	m := signpostMarshal(*t)
+	m.Metadata.Type = TypeSignpost
+	return json.Marshal(&m)
+}
+
+func (c *SignpostConfig) UnmarshalJSON(data []byte) error {
+	x := struct {
+		Trigger  *TypedObject `json:"trigger"`
+		Message  string       `json:"message,omitempty"`
+		Position Position     `json:"position"`
+	}{}
+
+	if err := json.Unmarshal(data, &x); err != nil {
+		return err
+	}
+
+	if x.Trigger != nil {
+		trigger, err := x.Trigger.ToComponent()
+		if err != nil {
+			return err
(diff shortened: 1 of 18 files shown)


wwitzel3:
Converted this to a draft PR. That is the best way to signal that something isn't ready for review.

pkg/view/component/signpost.go
```diff
@@ -0,0 +1,107 @@
...
+// MarshalJSON implements json.Marshaler
+func (t *Signpost) MarshalJSON() ([]byte, error) {
+	m := signpostMarshal(*t)
+	m.Metadata.Type = TypeSignpost
+	return json.Marshal(&m)
+}
+
+func (c *SignpostConfig) UnmarshalJSON(data []byte) error {
+	x := struct {
+		Component TypedObject `json:"component,omitempty"`
```
GuessWhoSamFoo, line 78: This needs to be a pointer otherwise it won't be empty if nil

pkg/view/component/signpost.go
```diff
@@ -0,0 +1,107 @@
...
+
+	if err := json.Unmarshal(data, &x); err != nil {
+		return err
+	}
+
+	component, err := x.Component.ToComponent()
+	if err != nil {
+		return err
+	}
+	trigger, err := x.Trigger.ToComponent()
```
GuessWhoSamFoo, line 94: This will panic if someone tries to pass a nil trigger.

web/src/app/modules/shared/components/presentation/signpost/signpost.component.ts
```diff
@@ -0,0 +1,44 @@
...
+  constructor(
+    private cdr: ChangeDetectorRef,
+    private readonly sanitizer: DomSanitizer
+  ) {
+    super();
+  }
+
+  update() {
+    this.directions.column = this.v.config.column || false;
+    this.order.reverse = this.v.config.reverse || false;
```
GuessWhoSamFoo, line 35: Aren't these false by default?

web/src/app/modules/shared/models/content.ts
```diff
@@ -599,3 +599,14 @@ export interface IconView extends View {
...
   };
 }
+
+export interface SignpostView extends View {
+  config: {
+    component: View;
+    trigger: View;
+    message: string;
+    column: false;
+    reverse: false;
```
GuessWhoSamFoo, line 609: These should be typed as boolean

GuessWhoSamFoo (changes requested):
How does a someone know if a component has a signpost?

The cursor changes to a pointer when hovered. Is this good enough?

https://clarity.design/angular-components/signpost/#behavior suggests also changing colors.

pkg/view/component/signpost.go
```diff
@@ -0,0 +1,107 @@
...
+)
+
+func NewSignpost(c Component, t Component, m string) *Signpost {
+	so := &Signpost{
+		Base: newBase(TypeSignpost, nil),
+		Config: SignpostConfig{
+			Component: c,
+			Trigger:   t,
+			Message:   m,
+			Reverse:   false,
```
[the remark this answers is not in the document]
lenriquez: Well, let say that you have a text and an icon, if you want the icon to be the trigger to show the signPost swapping will be a problem

pkg/view/component/signpost.go
lenriquez: Yes! trigger should never be empty otherwise will be imposible to trigger the signpost

web/src/app/modules/shared/components/presentation/signpost/signpost.component.ts
lenriquez: Yes! I will fix this

web/src/app/modules/shared/models/content.ts
lenriquez: oopss!!

web/src/app/modules/shared/components/presentation/signpost/signpost.component.ts
lenriquez: Actually, I think this could fail on storybook if values are not set properly

lenriquez:
> How does a someone know if a component has a signpost?
> 
> The cursor changes to a pointer when hovered. Is this good enough?
> 
> https://clarity.design/angular-components/signpost/#behavior suggests also changing colors.

Yes! you are right! but I wonder if I should ignore the given icon properties, for example if is a danger or warning it should show in gray and only red on hover?

web/src/app/modules/shared/models/content.ts
lenriquez: Done!

pkg/view/component/signpost.go
lenriquez: Done!

pkg/view/component/signpost.go
lenriquez: Done!

pkg/view/component/signpost.go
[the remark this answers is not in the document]
GuessWhoSamFoo: The only case I can think of where were need both a component and trigger is the text and icon combination. If this is true, maybe it is easier to make the message part of the icon component. Then this signpost component won't need a component and trigger with the reverse configuration going away.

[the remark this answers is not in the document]
lenriquez: I was thinking in the decorator pattern here (build large component base on small ones), I don't really know if is necessary but I have seen it on forms (first image), again I don't know if the icon should be with the label and we should use the column layout (second image), but it will be awesome if we can have this discussions during planning

image
image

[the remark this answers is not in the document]
lenriquez: done!

GuessWhoSamFoo (approved):
There could be scenarios in the future where the message is also a component. However, the current use case does not require this so we can leave this alone until needed.

PR #2222, opened 2021-03-25
Bump to Clarity 5

mklanjsek:
Updated our project to use Clarity `5.1.0` and fixed some compatibility issues:

- Fixed broken styles in storybook docs,
- Reworked left navigation colors to keep them as before,
- Added workaround for navigation [icon issue](https://github.com/vmware/clarity/issues/5644) (until it's fixed by clarity folks),
- Verified with @lenriquez that Icon component works properly,
- Also fixed annoying flicker in Storybook stories where components were initially updated in dark theme

Signed-off-by: Milan Klanjsek mklanjsek@pivotal.io


web/src/app/modules/shared/components/presentation/cytoscape/cytoscape.component.scss
```diff
@@ -16,7 +16,7 @@
   top: 0;
 }
 
-::ng-deep .label1 {
+::ng-deep div .label-header .label1 {
```
xtreme-vikram-yadav, line 19: Nitpicking here but wondering if `.label-header` can be applied inside element which is not a div since cytoscape can change the parent tag for the template. Can this be just `.label-header .label1` and still work fine?

mklanjsek: Right, good point - clarity 5 was overwriting these in storybook because they were not specific enough, but I went to other extreme and made therm probably too specific.

lenriquez:
Could you pls add the change log? :)

web/src/app/modules/shared/components/presentation/cytoscape/cytoscape.component.scss
mklanjsek: Changed it.

changelogs/unreleased/clarity5-mklanjsek
```diff
@@ -0,0 +1 @@
+Upgraded project to Clarity 5
```
GuessWhoSamFoo, line 1: Not blocking: Header spacing changes here (top -> 0.18, bottom -> current):

image

GuessWhoSamFoo (approved):

Issue #1513, opened 2020-10-16
Initialize plugins with current Octant state


scothis:
**Describe the problem/challenge you have**

When loading a plugin, it doesn't actually know what Octant's state is for values like the current namespace or filters. A plugin has to guess what the current values are until it observes an action change the value.

**Describe the solution you'd like**

Either an additional argument should be passed to plugins as they are initialized containing the current state, or plugins should have a way to lookup the current state on demand.

**Environment:**

- Octant version (use `octant version`): 0.16.1

wwitzel3:
I've been thinking about this recently, mainly for the JavaScript plugins, since I'm not sure how to do this in the Go plugins yet, having some lifecycle handlers here I think makes sense.

These would be called before the plugin was officially registered with Octant and started to receive requests:
- Setup can provide all of the current state like filters, namespace, context
- Start can be used as a handler to connect to external services, start watches (when we support that)

wwitzel3:
And the API fetching of current state for these values also makes sense. Expect an API endpoint in the short-term and some better lifecycle management in the long-term.

xtreme-vikram-yadav:
Forgot to mention but I picked this today.

wwitzel3:
We had a chat about this issue, we are going solve it in two ways:

- Request object will contain the current context
- Request object will contain the current namespace
- Request object will contain the current list of filters
- A new action will be added to allow plugins to watch for context change
- A new action will be added to allow plugins to watch for filter changes
- Watching for a namespace change already exists

This allows plugin authors to always pull the current state from request objects directly, without the need to make an API call back to Octant to get that information. No need to setup an action handler if you only care about the current state.

This allows plugin authors to react to those changes on demand if needed, for example, to clear/update local caches.

wwitzel3:
This doesn't solve the issue of the namespace being known during the "loading phase" .. @scothis do you have a problem with that given the described solution above? Or do you really want to have the current namespace before processing any requests?

scothis:
The current namespace (et al) on each request would address my concern.

wwitzel3:
fixed by #2244

PR #2244, opened 2021-03-29
Plugin request state

### LINKED ISSUE #1512: Allow plugins to get the current label filter and observe changes

Similar to how namespace changes can be observed in plugins to change the queries and data rendered, plugins also need to observe the label filters applied in Octant.

When a `addFilter`, `removeFilter` or `clearFilter` action is fired, a new action `setFilter` should be fired containing the updated aggregate filter state in its payload. There should also be a way to lookup the current filter state without observing changes.

**Environment:**

xtreme-vikram-yadav:
**What this PR does / why we need it**: This PR passes a shared partial octant state to the request objects passed to the plugins as arguments.

- Fixes #1512 #1513 

Some Design choices:
* The partial state is stored and passed via context. I felt conflicted about this choice as it didn't feel right to pass this data through context rather than a function argument. The reason context was picked to avoid updating multiple function arguments and interface implementation (which doesn't provide any meaning for the interfaces) along the multiple hops it takes for the state to go from state managers to plugins (for example: content_manager to a go plugin takes a lot of hops).
This way is a bit obscure but provides meaningful function params and data can be tracked with as much difficulty as a function argument would provide.

* clientId and shared plugin state is kept separate as clientId is used in other places and it doesn't update the existing protobufs.

* plugins get the state as `OState` property describing partial octant state.


pkg/action/actions.go
```diff
@@ -10,4 +10,14 @@ const (
 	// The ActionRequest.Payload for this action contains a single string entry `namespace` with a value
 	// of the new current namespace.
 	RequestSetNamespace = "action.octant.dev/setNamespace"
+
+	// RequestSteFilter is the action for when the filters are updated in Octant.
+	// The ActionRequest.Payload for this action contains an array of key/values representing
+	// the filters.
+	RequestSetFilter = "action.octant.dev/setFilter"
```
GuessWhoSamFoo, line 17: This can be consolidated with `RequestAddFilter`:

https://github.com/vmware-tanzu/octant/blob/9b658f64342cb6cb8388f689dc61a39cb79de4e5/internal/api/filter_manager.go#L22

pkg/action/actions.go
```diff
@@ -10,4 +10,14 @@ const (
 	// The ActionRequest.Payload for this action contains a single string entry `namespace` with a value
 	// of the new current namespace.
 	RequestSetNamespace = "action.octant.dev/setNamespace"
+
+	// RequestSteFilter is the action for when the filters are updated in Octant.
```
GuessWhoSamFoo, line 14: spelling

GuessWhoSamFoo (changes requested):

pkg/action/actions.go
xtreme-vikram-yadav: I had considered using `setFilter` but it seems like it could be used for any of the filter actions. Setting it for one will not read right. The only caveat if we go that route would be that the front end will be responsible for sending the appropriate aggregate filters list as a payload. Having a specific action name based on what it does reads a bit better. What are your thoughts about this?

web/src/stories/docs/plugins/GoPluginsIntro.story.mdx
```diff
@@ -115,6 +115,12 @@ log.Printf("octant-sample-plugin is starting")
 p.Serve()
 ```
 
+## Plugins with partial octant state
+All plugin implementations are passed a request parameter which makes a partial octant state available to the plugins.
+request.OState contains accessors for the current namespace and filters. The plugins can now respond to state updates by 
+either responding to specific actions (eg. action.octant.dev/setNamespace) or they can reflect the
+current state using the reflect.OState.
```
wwitzel3, line 122: s/reflect/request

wwitzel3:
I think I've sat on this long enough to decide I don't like the name `OState` .. it's too vague and isn't immediately obvious what it is. I think `ClientState` maybe a better choice.

Also, do we want to start to deprecate `ClientID` .. and move it in to `ClientState` ? So plugin authors get a `ClientState` object and it includes `ID`, `Filters`, `Namespace`.

I also think the state should include the current selected context.

xtreme-vikram-yadav:
Good catch. I totally missed including context. I had considered adding clientID to the state but I didn't add it as it is used in other places separately and it avoided removing clientID from protobuf for existing plugins which might be using it separately. But passing the reference to the state is still a valid solution. I'll update it :)

web/src/stories/docs/plugins/GoPluginsIntro.story.mdx
```diff
@@ -115,6 +115,12 @@ log.Printf("octant-sample-plugin is starting")
 p.Serve()
 ```
 
+## Plugins with partial octant state
+All plugin implementations are passed a request parameter which makes a partial octant state available to the plugins.
+request.OState contains accessors for the current namespace and filters. The plugins can now respond to state updates by
```
wwitzel3, line 120: s/OState/ClientState

wwitzel3 (approved):

GuessWhoSamFoo (approved):

GuessWhoSamFoo:
lets squash some of these commits

PR #2253, opened 2021-03-31
Render expression selector in selectors list

### LINKED ISSUE #2252: Expression filters are not rendered correctly
1. Apply a deployment that contains a `matchExpressions` selector. Example:
```yaml
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx
spec:
  selector:
    matchLabels:
      labelA: someLabelA
    matchExpressions:
      - key: labelB
        operator: In
        values:
          - someLabelB
  template:
    metadata:
      labels:
        labelA: someLabelA
        labelB: someLabelB
    spec:
      containers:
        - image: nginx
          name: nginx
```
2. Navigate to `Namespace Overview`, or any screen where deployments are listed
3. See values in the `Selector` column - they appear as: `labelA:someLabelA` and `labelB:`

Screenshot 2021-03-31 at 2 42 10 PM

1. The expression selector should be rendered as: `labelB In someLabelB`

It looks like the `selectors` type is being rendered as an `OverflowSelectorsComponent`, which is treating each selector as a label selector.

**Environment:**
Octant version: 0.18.0

xtreme-jon-ji:
**What this PR does / why we need it**: Fix expression selector rendering

- Fixes #2252

Couple of questions:
- Should `overflow-selectors` reuse the existing `label-selector` and `expression-selector` components? Wondering if `label-selector` portion of the `overflow-selector` was previously duplicated on purpose
- Is the design for `expression-selector` up-to-date? It doesn't use the same label styling as labels/label selector (see screenshot below)

Screenshot 2021-03-31 at 3 49 22 PM

Related issue:
- Selector filtering in `content-text-filter` treats selectors as label selectors, with only `key` and `value` fields. Thus, the filter does not search within expression operator/values. Should this be fixed alongside this PR, or should I raise a separate issue?

### DIFF
--- a/web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
+++ b/web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
@@ -6,7 +6,7 @@
       <app-octant-tooltip *ngSwitchCase="'labelSelector'" tooltipText="{{ selector.config.key }}:{{ selector.config.value }}">
         <app-view-label-selector [view]="selector"></app-view-label-selector>
       </app-octant-tooltip>
-      <app-octant-tooltip *ngSwitchCase="'expressionSelector'" tooltipText="{{ selector.config.key }} {{ selector.config.operator }} {{ selector.config.values?.join('|') }}">
+      <app-octant-tooltip *ngSwitchCase="'expressionSelector'" tooltipText="{{ selector.config.key }} {{ selector.config.operator }} {{ selector.config.values | join: ',' }}">
         <app-view-expression-selector [view]="selector"></app-view-expression-selector>
       </app-octant-tooltip>
     </ng-container>
--- a/web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
+++ b/web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
@@ -1,12 +1,13 @@
 <div class="selectors-wrapper" *ngIf="selectors">
   <div class="selectors-container">
     <ng-container
-      *ngFor="let selector of showSelectors; trackBy: trackByIdentity"
+      *ngFor="let selector of showSelectors; trackBy: trackByIdentity" [ngSwitch]="selector.metadata.type"
     >
-      <app-octant-tooltip tooltipText="{{ selector.config.key }}:{{ selector.config.value }}">
-        <span class="label label-orange">
-          {{ selector.config.key }}:{{ selector.config.value }}
-        </span>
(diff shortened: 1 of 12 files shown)


web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
```diff
@@ -1,12 +1,13 @@
...
+      *ngFor="let selector of showSelectors; trackBy: trackByIdentity" [ngSwitch]="selector.metadata.type"
     >
-      <app-octant-tooltip tooltipText="{{ selector.config.key }}:{{ selector.config.value }}">
-        <span class="label label-orange">
-          {{ selector.config.key }}:{{ selector.config.value }}
-        </span>
+      <app-octant-tooltip *ngSwitchCase="'labelSelector'" tooltipText="{{ selector.config.key }}:{{ selector.config.value }}">
+        <app-view-label-selector [view]="selector"></app-view-label-selector>
+      </app-octant-tooltip>
+      <app-octant-tooltip *ngSwitchCase="'expressionSelector'" tooltipText="{{ selector.config.key }} {{ selector.config.operator }} {{ selector.config.values?.join(',') }}">
```
GuessWhoSamFoo, line 9: I'd prefer to see this join as a pipe

GuessWhoSamFoo (changes requested):

web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
```diff
@@ -1,12 +1,13 @@
...
+        <app-view-expression-selector [view]="selector"></app-view-expression-selector>
```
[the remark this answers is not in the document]
xtreme-jon-ji: Screenshot 2021-04-01 at 11 47 28 AM

Updated to use purple label and pipe

web/src/app/modules/shared/components/presentation/overflow-selectors/overflow-selectors.component.html
GuessWhoSamFoo: Sorry if I was unclear, this was meant to be https://angular.io/guide/pipes#using-a-pipe-in-a-template

The reason why we do this is because a pure pipe ensures Angular only re-renders if inputs change rather than relying on the component's change detection strategy.

xtreme-jon-ji: My bad, I didn't know about Angular pipes. Thanks for teaching me something new! I've added a `JoinPipe`, and reverted the delimiter to be a comma

GuessWhoSamFoo (approved):
lgtm, thanks for the fix

PR #2254, opened 2021-03-31
Refactor forms to improve re-usability

### LINKED ISSUE #1504: Normalize form capabilities

lenriquez:

Refactor Form code does not have to be repeated on both the stepper and the form component

- Fixes #1504

**Solution 1**
- Add dynamically stepper directives and html necessary
**Why did not work**
- Adding dynamically directives is hard and is not implemented in angular by default
- Form will share two responsibilities: Create a Form and a Stepper

**Solution 2**
- Create a dynamically load form as is done on view-container.component or as is described here https://angular.io/guide/dynamic-component-loader
**Why did not work**
Most of the "Form inputs" are not components but HTML that use directives, doing this does not help

**Solution 3**
Just split the code
 - New component was created contains all the logic for the template
 - form-helper was created, so it can be use by any class and no code is repeated, decided to use composition over inheritance
 
**Release note**:
```
release-note
```


web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
```diff
@@ -0,0 +1,111 @@
...
+          clrInput
+          type="text"
+          [formControlName]="field.name"
+          [placeholder]="field.placeholder ? field.placeholder : ''"
+        />
+        <clr-control-error>{{ field.error }}</clr-control-error>
+      </clr-input-container>
+    </ng-container>
+
+    <ng-container *ngSwitchCase="'number'">
```
GuessWhoSamFoo, line 55: If the default value is used instead of an updated value, the return type is a string. A form type number should always return a number.

GuessWhoSamFoo (changes requested):

web/src/app/modules/shared/components/form-view-container/form-view-container.component.ts
```diff
@@ -0,0 +1,41 @@
...
+    } else {
+      this.formArray.controls.forEach((fc: FormControl, index: number) => {
+        if (fc.value === event.target.value) {
+          this.formArray.removeAt(index);
+        }
+      });
+    }
+  }
+
+  trackByFn(index, _) {
```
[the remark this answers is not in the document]
lenriquez: I see this in many places 
image
Are you saying that I should put it on the AbstractViewComponent class or create a util file?

web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
lenriquez: I am not completely sure what you mean but I was thinking in something like this 
`
  transformValue(field): any {
    if (field.type === 'number') {
      const value = +field.value;
      return Number.isNaN(value) ? 0 : value;
    }
  }`
this function will be called during form creation

lenriquez: And if is empty string should I return 0 or nil?

web/src/app/modules/shared/components/form-view-container/form-view-container.component.ts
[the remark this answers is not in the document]
lenriquez: It was this file trackByIndex.ts

web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
lenriquez: it was decided to use nil

lenriquez: Done!

web/src/app/modules/shared/models/form-helper.ts
```diff
@@ -0,0 +1,84 @@
...
+// Class responsible to create a Form Group and add Validations Functions to form control
+export class FormHelper {
+  createFromGroup(form, formBuilder) {
+    if (!form) {
+      return;
+    }
+
+    const controls: { [name: string]: any } = {};
+    form.fields.forEach(field => {
+      // createControlGroups
```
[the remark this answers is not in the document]
lenriquez: Done!

GuessWhoSamFoo (approved):

PR #2255, opened 2021-03-31
Created ButtonComponent and extended LinkComponent to receive components to be wrapped as links

### LINKED ISSUE #2183: Buttons should be able to link

ftovaro:
Signed-off-by: ftovaro <f.tovar12@gmail.com>

It solves two things:

1. We can now have buttons without a button group
2. We can pass components to a LinkComponent to use it as a link

- Fixes #2183

This is how the button looks like, here is being used as a link:

Screen Shot 2021-04-05 at 2 33 04 PM

This is the HTML generated for the button inside a LinkComponent:
```
<a target="_blank" href="http://www.google.com" class="ng-star-inserted">
   <app-view-container class="component" ...>
      <app-single-button class="ng-star-inserted">
         <button class="btn btn-outline btn-sm"> single button in link</button>
      </app-single-button>
   </app-view-container>
</a>
```

This is how you can pass a component to a LinkComponent:
```
btn := component.NewButton("button in link", action.Payload{})
btnLink := component.NewLink("title", "value","http://www.google.com", component.WithComponent(btn))
```

**Release note**:
```
release-note
```

### DIFF
--- a/internal/printer/service_test.go
+++ b/internal/printer/service_test.go
@@ -192,7 +192,7 @@ func Test_ServiceConfiguration(t *testing.T) {
 
 	startPortForwardingButtonGroup := component.NewButtonGroup()
 	startPortForwardingButtonGroup.Config = component.ButtonGroupConfig{Buttons: []component.Button{
-		component.NewButton("Start port forward", action.Payload{
+		*component.NewButton("Start port forward", action.Payload{
 			"action":     "overview/startPortForward",
 			"apiVersion": validService.APIVersion,
 			"kind":       validService.Kind,
@@ -204,7 +204,7 @@ func Test_ServiceConfiguration(t *testing.T) {
 
 	stopPortForwardingButtonGroup := component.NewButtonGroup()
 	stopPortForwardingButtonGroup.Config = component.ButtonGroupConfig{Buttons: []component.Button{
-		component.NewButton("Stop port forward", action.Payload{
+		*component.NewButton("Stop port forward", action.Payload{
 			"action": "overview/stopPortForward",
 			"id":     "an-id",
 		}),
--- a/pkg/view/component/base.go
+++ b/pkg/view/component/base.go
@@ -90,6 +90,8 @@ const (
 	TypeIcon = "icon"
 	// TypeSignpost is a SignPost component.
 	TypeSignpost = "signpost"
+	// TypeButton is a Button component.
+	TypeButton = "button"
 )
 
 // Base is an abstract base for components..
--- a/pkg/view/component/button.go
+++ b/pkg/view/component/button.go
@@ -1,10 +1,14 @@
+/*
+Copyright (c) 2021 the Octant contributors. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
 package component
 
 import (
-	"github.com/vmware-tanzu/octant/internal/util/json"
-
 	"github.com/pkg/errors"
 
+	"github.com/vmware-tanzu/octant/internal/util/json"
 	"github.com/vmware-tanzu/octant/pkg/action"
 )
 
@@ -20,7 +24,7 @@ type ButtonOption func(button *Button)
 // WithButtonConfirmation configured a button with a confirmation.
 func WithButtonConfirmation(title, body string) ButtonOption {
 	return func(button *Button) {
-		button.Confirmation = &Confirmation{
+		button.Config.Confirmation = &Confirmation{
 			Title: title,
 			Body:  body,
 		}
@@ -30,11 +34,15 @@ func WithButtonConfirmation(title, body string) ButtonOption {
 // WithModal configures a button to open a modal
 func WithModal(modal *Modal) ButtonOption {
 	return func(button *Button) {
-		button.Modal = modal
+		button.Config.Modal = modal
 	}
 }
 
-func (b *Button) UnmarshalJSON(data []byte) error {
+var _ Component = (*Button)(nil)
+
+type buttonMarshal Button
+
+func (b *ButtonConfig) UnmarshalJSON(data []byte) error {
 	x := struct {
 		Name         string         `json:"name"`
 		Payload      action.Payload `json:"payload"`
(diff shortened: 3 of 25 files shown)


pkg/view/component/base.go
```diff
@@ -88,6 +88,8 @@ const (
 	TypeYAML = "yaml"
 	// TypeIcon is a Icon component.
 	TypeIcon = "icon"
+	// TypeSingleButton is a SingleButton component.
+	TypeSingleButton = "singleButton"
```
wwitzel3, line 92: Just `button`. `ButtonGroup` is made up of `Button` componoents. A single button should be just `button` 

- `NewButton`
- `TypeButton = "button"`

wwitzel3 (changes requested):

pkg/view/component/single_button.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"github.com/pkg/errors"
+	"github.com/vmware-tanzu/octant/internal/util/json"
+	"github.com/vmware-tanzu/octant/pkg/action"
+)
+
+// SingleButton is a component for a SingleButton
+// +octant:component
+type SingleButton struct {
```
wwitzel3, line 16: Just `Button`

pkg/view/component/base.go
ftovaro: Got it, I'm gonna do the fix right now

pkg/view/component/link.go
```diff
@@ -95,6 +106,11 @@ func (t *Link) SetStatus(status TextStatus, detail Component) {
 	t.Config.StatusDetail = detail
 }
 
+// SetComponent sets the component to be wrapped by the Link
```
wwitzel3, line 109: Let's also add a `WithComponent` that can be passed as a `LinkOption`

wwitzel3 (changes requested):

pkg/view/component/base.go
ftovaro: @wwitzel3 now I remember why I didn't call the component just `button` (I thought about it too),

The GroupButtonComponent files are already called for example `button.go`, so that's why I came up with the idea of calling it `SingleButton` to avoid collisions,  so I'm gonna try to make the change of the GroupButtonComponent files and renamed them to `button_group.go`  so we can have a better naming.

pkg/view/component/single_button.go
ftovaro: It collides with the definition on https://github.com/vmware-tanzu/octant/blob/master/pkg/view/component/button.go#L70
So, to avoid that, I'm gonna rename in that file the struct to ButtonElement to be able to rename this one to just `Button`

ftovaro: So, I checking this change and I think it can be a little aggressive because in line https://github.com/vmware-tanzu/octant/blob/master/pkg/view/component/button.go#L94 we define the array of `Buttons`, if I change the structure to be `ButtonElement` the plugins are creating buttons groups like this are gonna break: 

```
					ButtonGroup: &ButtonGroup{
						Base: Base{},
						Config: ButtonGroupConfig{
							Buttons: []Button{{
								Name: "test",
							}},
						},
					},
```

wwitzel3: That would still be `[]Button` .. the `Button` struct itself, now needs to inheirt from base and you'll want a `ButtonConfig` type to represent the button options.

wwitzel3: Similar to this, https://gist.github.com/wwitzel3/82fb4d1e0b6dad1fdb9bb204ff03214c

wwitzel3: If we need to break existing APIs to ensure we have the correct ones as we move towards 1.0, then that is what we have to do. `ButtonComponent` is what we should have. `ButtonGroup.Buttons` should be made up of those components.

web/src/app/modules/shared/components/presentation/button-group/button-group.component.html
```diff
@@ -1,25 +1,34 @@
...
 <clr-button-group *ngIf="v.config.buttons" class="btn-primary">
   <clr-button
     *ngFor="let button of v.config.buttons; trackBy: trackByFn"
     class="{{ class }}"
-    (click)="onClick(button.payload, button.confirmation, button.modal)"
+    (click)="onClick(button.config.payload, button.config.confirmation, button.config.modal)"
   >
-    {{ button.name }}
+    {{ button.config.name }}
+    xxxx
```
lenriquez, line 10: This xxxx should be here?

web/src/app/modules/shared/components/presentation/button-group/button-group.component.html
```diff
@@ -1,25 +1,34 @@
...
+
+    xxx
```
lenriquez, line 12: Same here?

pkg/view/component/link.go
ftovaro: Added at line https://github.com/vmware-tanzu/octant/pull/2255/files#diff-e4b9f4aec321f580471b37d920edfa5164cc21ea234e4ff7395f96926480bc64R139

Here is an example of how to use it:

```
btn := component.NewButton("single button in link", action.Payload{})
btnLink := component.NewLink("title", "value", requestPath, component.WithComponent(btn))
```

pkg/view/component/button.go
```diff
@@ -1,40 +1,64 @@
...
+	Payload      action.Payload `json:"payload"`
+	Confirmation *Confirmation  `json:"confirmation,omitempty"`
+	Modal        Component      `json:"modal,omitempty"`
 }
 
 // ButtonOption is a function for configuring a Button.
 type ButtonOption func(button *Button)
 
-// WithButtonConfirmation configured a button with a confirmation.
-func WithButtonConfirmation(title, body string) ButtonOption {
```
wwitzel3, line 21: I don't think we want to remove this?

pkg/view/component/button.go
```diff
@@ -1,40 +1,64 @@
...
+		Base: newBase(TypeButton, nil),
+		Config: ButtonConfig{
+			Name:    name,
+			Payload: payload,
+		},
 	}
-}
 
-// WithModal configures a button to open a modal
-func WithModal(modal *Modal) ButtonOption {
```
wwitzel3, line 31: Same here, add this back.

pkg/view/component/button_group.go
```diff
@@ -0,0 +1,91 @@
...
+	return func(button *Button) {
+		button.Config.Confirmation = &Confirmation{
+			Title: title,
+			Body:  body,
+		}
+	}
+}
+
+// WithModal configures a button to open a modal
+func WithModal(modal *Modal) ButtonOption {
```
wwitzel3, line 26: This belongs in button.go

pkg/view/component/button_group.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+// Confirmation is configuration for a confirmation dialog.
+type Confirmation struct {
+	Title string `json:"title"`
+	Body  string `json:"body"`
+}
+
+// WithButtonConfirmation configured a button with a confirmation.
+func WithButtonConfirmation(title, body string) ButtonOption {
```
wwitzel3, line 16: This belongs in button.go

pkg/view/component/port.go
```diff
@@ -135,7 +135,7 @@ func describeButton(namespace, apiVersion, kind, name string, port int, pfs Port
 	}
 
 	buttonGroup.AddButton(
-		NewButton(buttonText, action.CreatePayload(actionName, payload)),
+		*NewButton(buttonText, action.CreatePayload(actionName, payload)),
```
wwitzel3, line 138: `NewButton` already returns a pointer to the button (`*Button` is the return type), you shouldn't need the `*`

pkg/view/component/table.go
```diff
@@ -287,7 +287,7 @@ func (t *Table) AddButton(name string, payload action.Payload, buttonOptions ...
 		t.Config.ButtonGroup = NewButtonGroup()
 	}
 	button := NewButton(name, payload, buttonOptions...)
-	t.Config.ButtonGroup.AddButton(button)
+	t.Config.ButtonGroup.AddButton(*button)
```
wwitzel3, line 290: This is an inconsistent API experience. The caller should not have to have knowledge of how to transform the button to pass it in to `AddButton` .. `NewButton` should return the type that `AddButton` expects and then we can remove the `*`

pkg/view/flexlayout/flexlayout.go
```diff
@@ -33,7 +33,7 @@ func (fl *FlexLayout) AddSection() *Section {
 // AddButton adds a button the button group for a flex layout.
 func (fl *FlexLayout) AddButton(name string, payload action.Payload, buttonOptions ...component.ButtonOption) {
 	button := component.NewButton(name, payload, buttonOptions...)
-	fl.buttonGroup.AddButton(button)
+	fl.buttonGroup.AddButton(*button)
```
wwitzel3, line 36: Same comment as other places, make `AddButton` accpet the return of `NewButton`

pkg/view/flexlayout/flexlayout_test.go
```diff
@@ -70,7 +70,7 @@ func TestFlexLayout_add_button(t *testing.T) {
 	got := fl.ToComponent("Title")
 	expected := component.NewFlexLayout("Title")
 	buttonGroup := component.NewButtonGroup()
-	buttonGroup.AddButton(component.NewButton("button", action.Payload{}))
+	buttonGroup.AddButton(*component.NewButton("button", action.Payload{}))
```
wwitzel3, line 73: Same comment as other places, make AddButton accpet the return of NewButton

This call should be: `buttonGroup.AddButton(component.NewButton("button", action.Payload{}))`

wwitzel3 (changes requested):

pkg/view/component/table.go
ftovaro: You are absolutely right, I just made the update for the `AddButton` method

internal/printer/service_test.go
```diff
@@ -204,7 +204,7 @@ func Test_ServiceConfiguration(t *testing.T) {
 
 	stopPortForwardingButtonGroup := component.NewButtonGroup()
 	stopPortForwardingButtonGroup.Config = component.ButtonGroupConfig{Buttons: []component.Button{
-		component.NewButton("Stop port forward", action.Payload{
+		*component.NewButton("Stop port forward", action.Payload{
```
wwitzel3, line 207: Should not have to use `*`

ftovaro: Not sure about this one 'cause we are creating and array of objects, so you need to pass the references of the objects right?

ftovaro: So, I was checking about this and found that I can change

```
type ButtonGroupConfig struct {
	// Buttons are buttons in the group.
	Buttons []Button `json:"buttons"`
}
```

To be

```
type ButtonGroupConfig struct {
	// Buttons are buttons in the group.
	Buttons []*Button `json:"buttons"`
}
```

But, with that change, it means that the line you commented would need this change:

```
stopPortForwardingButtonGroup.Config = component.ButtonGroupConfig{Buttons: []*component.Button{
              component.NewButton("Stop port forward", action.Payload{

```

Because now the array is going to be of references.

So before making the change, just want to make sure that is what we want.

wwitzel3: all right this is fine for now, we can leave this alone

wwitzel3 (approved):

PR #2257, opened 2021-03-31
Expose Delete and Create object store calls for Go Plugins

### LINKED ISSUE #1599: Provide APIs to create and delete resources for plugins to use
I would like to switch to DashboardClient for antrea-octant-plugin, so that the plugin can pick up context switches from octant dashboard. My plugin needs to create and delete specific CRDs which can be supported by plugin self-generated client using client-go and kubeconfig previously. However, there are no APIs for create and delete resources defined in Dashboard interface (https://github.com/vmware-tanzu/octant/blob/master/pkg/plugin/service/dashboard.go#L16) which I can invoke. It would be great if octant can provide create and delete resources APIs for plugin to use.

Add Create and Delete APIs to Dashboard interface with their implementations, so that plugins can use to create and delete resources of their own.

**Environment:**

- Kubernetes version (use `kubectl version`): v1.18.5
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):Ubuntu 18.04.3

xtreme-vikram-yadav:
**What this PR does / why we need it**: This PR creates/exposes Create and Delete Calls for Go plugins to interact with the object store.

- Fixes #1599 

Create Call was already available(https://github.com/vmware-tanzu/octant/blob/9b658f64342cb6cb8388f689dc61a39cb79de4e5/pkg/plugin/api/server.go#L270). 
It just wasn't exposed via Dashboard interface (https://github.com/vmware-tanzu/octant/blob/master/pkg/plugin/service/dashboard.go).


GuessWhoSamFoo (approved):
lgtm

PR #2273, opened 2021-04-05
Bumped angular, ansi_up

mklanjsek:
Upgraded client side dependencies:
- angular from `11.2.3` to `11.2.8`,
- ansi_up from `4.0.4` to `5.0.0` to address the cross-site scripting vulnerability


web/src/app/modules/shared/pipes/ansiPipe/ansi.pipe.ts
```diff
@@ -15,9 +15,7 @@ import { default as AnsiUp } from 'ansi_up';
 export class AnsiPipe implements PipeTransform {
   ansiUp = new AnsiUp();
 
-  constructor(private sanitizer: DomSanitizer) {
-    this.ansiUp.escape_for_html = false;
```
wwitzel3, line 19: Were we depending on this behavior anywhere? I'm wondering why we were explicitly setting it to false.

wwitzel3 (approved):
I have one question, but as long as we weren't depending on the behavior enabled by `escape_for_html` this is good to merge.

web/src/app/modules/shared/pipes/ansiPipe/ansi.pipe.ts
mklanjsek: @wwitzel3 That is now enforced to be false to prevent security risks and therefore removed from the interface. So, the internal behavior is pretty much same as before. I also tested manually ansi codes work as before.

Issue #1925, opened 2021-02-01
npm ERR! cb() never called!


GuessWhoSamFoo:
xref: https://github.com/vmware-tanzu/octant/runs/1807187681?check_suite_focus=true#step:4:6

Likely has something to do with invalidating the npm cache

PR #2275, opened 2021-04-06
Update CI to leverage build.go commands

### LINKED ISSUE #2211: npm run electron:build fails
After merging in the latest changes for the Resource Viewer, running `npm run electron:build` fails because it directly calls `ng build` and does not run the `pre-install` and `post-install` steps that happen when running `npm install`.

Running `go run build.go web-build` also fails for this same reason.

A quick fix is to ensure `npm install` is run before the main `build` command is run, but also, I would like us to audit our usage of build.go and our pipeline to make sure we aren't duplicating efforts and preferring and using `build.go` when we can.

xtreme-vikram-yadav:
**What this PR does / why we need it**: This PR updates CI to make use of build.go commands where it can. Commands now setup their own dependencies before running.

- Fixes #2211 

A few commands now setup their own dependencies before running so there is no additional setup step needed on CI. This means that in a couple cases we run duplicate setup commands. This seems like a good tradeoff as the commands do not need external setup and any subsequent setup is much faster.


xtreme-vikram-yadav:
build.go is a collection of commands without much structure and logic. There doesn't seem like a good way to test without increasing the structural complexity. The correctness is tested by running these commands in CI and development if testing sequential execution is the goal.

GuessWhoSamFoo:
build.go is meant to be an OS-independent stand-in for makefiles and should only contain convenience methods for development. Once we move to electron only releases, a lot of the complexity will go away.

.github/workflows/preflight-checks.yaml
```diff
@@ -86,16 +84,6 @@ jobs:
         with:
           go-version: ${{ matrix.go-version }}
       - uses: actions/checkout@v2
-      - name: Build node modules
-        run: |
-          cd web
-          npm ci --prefer-offline --no-audit
-          npm run-script build
-        shell: bash
-      - uses: actions/upload-artifact@v2
```
GuessWhoSamFoo, line 95: Why is this removed? https://github.com/vmware-tanzu/octant/blob/af6ee1bcee8ab61f87f9949982e1fe308650c8ed/.github/workflows/preflight-checks.yaml#L140-L143 This step will download nothing if it is a release

GuessWhoSamFoo (changes requested):

.github/workflows/preflight-checks.yaml
xtreme-vikram-yadav: Oh It's my bad. I didn't realize that this is being used across jobs. I see the same problem with nightly.yml then which has goreleaser using it. I'll revert it.

GuessWhoSamFoo:
Added `npm cache verify` as an attempt to address #1925

GuessWhoSamFoo (approved):

PR #2276, opened 2021-04-06
Add link generator to DashboardClient

### LINKED ISSUE #1444: In Go extensions API provide utility method to create the Link ref for the specified Kubernetes object

In our Starboard plugin, which is implemented in Go, we are displaying custom resources that have owner references set. To create the UI Link components we'd have to reverse-engineer the logic to build such URL refs. For example, we'd like to link to a Deployment or ReplicaSet with the given name in the specified namespace.

plugin_go_link_builder

Go API method to construct `Link` components for the specified `ObjectMeta` and/or `OwnerReference`. A similar method exists for JS/TS: https://github.com/vmware-tanzu/octant/blob/master/pkg/plugin/javascript/dashboard_ref_path.go

N/A

**Environment:**

GuessWhoSamFoo:

Usage: `request.DashboardClient.CreateLink(request.Context(), key)`

 - Returns a link response that has a link component with the object's name as the link text as default.

- Fixes #1444


wwitzel3 (approved):
Thank you for checking this one off the list.

wwitzel3:
Can you create an issue to add this to an example plugin as well? I think showing the usage of this will be useful to plugin authors.

GuessWhoSamFoo:
I'm inclined to say storybook should be for component rather than API documentation. We can add an example plugin and create a page for documenting what DashboardClient can do outside of code comments.

wwitzel3:
Where do you see us moving these to? https://reference.octant.dev/?path=/docs/docs-plugins-2-go-plugins--page

wwitzel3:
My thought was that having the "how to" plugin guides and component reference in the same place is useful

GuessWhoSamFoo:
That's fine where it is. I'll add sample code in https://github.com/vmware-tanzu/octant-example-plugins

Issue #2213, opened 2021-03-24
Adapt Icon Component for Clarity 5


lenriquez:
**Describe the problem/challenge you have**
[A description of the current limitation/problem/challenge that you are experiencing.]
Icon component at this moment use clarity 4 and there are some differences between versions
**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
To Adapt this change is necessary to removed the mapping and make sure the UI has the right [attr] the component should continue working as it is 
Also change documentation on story book
**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Octant version (use `octant version`): 
- Kubernetes version (use `kubectl version`):
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):

PR #2286, opened 2021-04-09
Use cds-icon instead of clr-icon for icon component

GuessWhoSamFoo:
`clr-icon` has reached deprecation in clarity 4 and will receive no future updates for 5+.

This PR changes the icon component to use `cds-icon`. It adds helper methods with descriptions for various custom behaviors. Small storybook inconsistencies are addressed.

It also removes a lot of custom logic to generate correct styling based on class attributes. This should be part of 0.19 as many new icons are now available.

- Partially addresses #2213 

Internal usage of `clr-icon` within other components should be updated as well, but this does not need to be tied to a release.


web/src/stories/icon.stories.mdx
```diff
@@ -20,7 +20,7 @@ c.Config.Badge = component.BadgeDanger
 <Meta title="Components/Icon" component={IconComponent} />
 
 <Canvas withToolbar>
```
mklanjsek, line 22: Not sure the last `<p>` above is accurate any more, can't we remove Clarity 4 reference (same for signpost)?

web/src/stories/icon.stories.mdx
```diff
@@ -29,14 +29,13 @@ c.Config.Badge = component.BadgeDanger
...
           config: {
             shape: "user",
-            size: "65",
+            size: "md",
             direction: "up",
             flip: "horizontal",
             solid: true,
             status: "danger",
             inverse: false,
             badge: "danger",
```
mklanjsek, line 38: Can you add badge color and label, we always try to include all properties (same for signpost).

mklanjsek (changes requested):
Looks good, just a few small tweaks. Can please create an issue to address switch for internal icons, would be good to have them in sync ASAP.

web/src/app/modules/shared/models/content.ts
```diff
@@ -591,6 +591,8 @@ export interface IconView extends View {
     inverse: boolean;
     badge: string;
     color: string;
+    badgeColor: string;
+    label: string;
```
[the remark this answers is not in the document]
GuessWhoSamFoo: Go returns empty strings, optional wouldn't be meaningful here

GuessWhoSamFoo:
Fixed improper setting of aria-label and loading of `--color` and `--badge-color` styles. Removed references to clarity 4

mklanjsek (approved):
🚢

PR #2325, opened 2021-04-13
Expose client state as an interface from service

### LINKED ISSUE #2305: ClientState uses a private struct
ClientState is a struct out of internal and makes it non-mockable for testing.

Use an interface to expose this to plugin authors from the service package.

```
type ClientState interface {
	ClientID() string
	Filters() []Filter
	Namespace() string
	ContextName() string
}
```

xtreme-vikram-yadav:
**What this PR does / why we need it**: This PR introduces client state as an interface to make it easy for plugin authors writing tests.

- Fixes #2305 

**Special notes for your reviewer**: There are many considerations when implementing this because of the way it affects the way json data is handled between octant and plugins and how data is exposed to the JS plugins.

Some design decisions made:
* the client state implementation resides in the original octant context file to keep the dependency flow from octant to service (It could be moved to its own file but I'm not sure what to call it yet).
* the client state is kept private and can be accessed only via accessor methods from client state interface. This avoids creation of multiple points of access both in JS and GO plugins.
* Because client state is private, the JSON marshalling/unmarshalling is handled by custom methods.

Alternatives to this implementation:
All the above problems/workarounds can be addressed by having the client state interface with get methods:
```
type ClientState interface {
	GetClientID() string
	GetFilters() []Filter
	GetNamespace() string
	GetContextName() string
}
```
 which will be implemented by https://github.com/vmware-tanzu/octant/blob/11bfafcac30a4a78c07ee6b7196e212f262d7f04/internal/context/context.go#L30.

Although it improves the code readability, it now exposes two ways to access the state (via fields and getter methods). I doubt anyone will ever use the getters in this case which makes the interface usable only for tests. 
The PR implementation introduces some complexity but exposes a consistent way to access state from both types of plugins.


internal/context/context.go
```diff
@@ -27,20 +28,72 @@ type OctantClientState string
...
+
+type jsonClientState struct {
 	ClientID    string          `json:"clientID"`
 	Filters     []octant.Filter `json:"filters"`
 	Namespace   string          `json:"namespace"`
 	ContextName string          `json:"contextName"`
 }
 
-func WithClientState(ctx context.Context, state ClientState) context.Context {
+type clientState struct {
```
ftovaro, line 56: Just out of curiosity, why is the `client` in `clientCase` lowercase here when it was previously uppercase

ftovaro:
So if a I understand correctly what you did here, you created some methods to access the attributes of the ClientState in order to allow plugins authors to access those attributes without exposing the ClientState itself, right?

internal/context/context.go
wwitzel3: This is a go convention to make something private, this helps us enforce folks using the interference and helper methods instead of the struct directly

wwitzel3:
Tom looked at this and was still having trouble because Filter type is internal. I think we need to lift that up as well.

xtreme-vikram-yadav:
> Tom looked at this and was still having trouble because Filter type is internal. I think we need to lift that up as well.

Ah right. I'll ix it.

internal/context/context.go
ftovaro: Got it, thanks!

GuessWhoSamFoo (approved):

PR #2328, opened 2021-04-14
Convert clr-modal to cds-modal

GuessWhoSamFoo:
**What this PR does / why we need it**:
Shifts usage of clr-modal to cds-modal. Minor style adjustments and additional accessibility features are gained from this update.

**Which issue(s) this PR fixes**
- See https://kubernetes.slack.com/archives/CM37M9FCG/p1618414173115100

**Special notes for your reviewer**:
Places of interest;
 - Build and release information on the top right menu
 - Preferences
 - Quick switcher
 - Confirmation dialog when deleting an object (from datagrids and object summary pages)
 - a modal opened by a button

Signed-off-by: Sam Foo <foos@vmware.com>

PR #2330, opened 2021-04-15
Fix Modal Cursor

### LINKED ISSUE #1952: Cursor should not always be pointer when modal is open
Seems to be affecting electron only, and not the preferences modal

xtreme-vikram-yadav:
**What this PR does / why we need it**: The cursor is not always a pointer.

- Fixes #1952


xtreme-vikram-yadav:
Keeping it as draft until https://github.com/vmware-tanzu/octant/pull/2328/ is merged.

PR #2350, opened 2021-04-21
Add tooltip to icon component

### LINKED ISSUE #2307: Extend IconComponent to support Tooltip
Teams want to be able to use tooltips to help onboard and inform users about taxonomy and topology

Implement the tooltips component from clarity

Clarity docs: https://clarity.design/angular-components/tooltip/#tooltip-right

GuessWhoSamFoo:
Adds a tooltip to icons. Only added to icons as it seems to be the few components where adding a tooltip makes sense over a signpost.

- Fixes #2307 

### DIFF
--- a/web/src/app/modules/shared/components/presentation/icon/icon.component.ts
+++ b/web/src/app/modules/shared/components/presentation/icon/icon.component.ts
@@ -16,7 +16,7 @@ import {
   loadTechnologyIconSet,
   loadChartIconSet,
 } from '@cds/core/icon';
-import { IconView } from '../../../models/content';
+import { IconView, Tooltip } from '../../../models/content';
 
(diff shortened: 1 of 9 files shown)


web/src/app/modules/shared/components/presentation/icon/icon.component.ts
```diff
@@ -51,6 +53,17 @@ export class IconComponent extends AbstractViewComponent<IconView> {
 
   protected update(): void {
     const view = this.v;
+    this.tooltip = view.config.tooltip;
```
lenriquez, line 56: This function is already long it will be nice to have this functionality in another function

lenriquez (approved):
It looks good to me!

PR #2331, opened 2021-04-15
Website migration to Hugo

jonasrosland:
**What this PR does / why we need it**:
This will migrate the existing site from Jekyll to Hugo

**Which issue(s) this PR fixes**
- Fixes #

**Special notes for your reviewer**:

**Release note**:
```
release-note
```

PR #2326, opened 2021-04-13
Move contributors page to community, remove from home page.

lenriquez:
Signed-off-by: Luis Enriquez <felipe096@gmail.com>

**What this PR does / why we need it**:
I am a new approver!!!

**Which issue(s) this PR fixes**
- Fixes #2312

**Special notes for your reviewer**:

image

**Release note**:
```
release-note
```

PR #2374, opened 2021-04-26
Website revamp to Hugo

GuessWhoSamFoo:
Supersedes #2331 

TODO:
 - [x] Re-add #2326 
 - [x] netlify.toml to generate site preview
 - [ ] General audit of changes


site/config.yaml
```diff
@@ -0,0 +1,50 @@
...
+  use_advanced_docs: false
+  docs_right_sidebar: true
+  docs_search: false
+  docs_search_index_name: index_name
+  docs_search_api_key: api_key
+  docs_versioning: true
+  docs_latest: v0.40.0
+  docs_versions:
+    - v0.30.0
+    - v0.40.0
```
jonasrosland, line 38: This will make the /docs point to /docs/v0.40.0 (this is a leftover from the template), which doesn't exist.
You can replace this with something like "latest", and add a "latest" folder under /docs/content, and a Markdown file with the same brief message as is currently being used.

jonasrosland (changes requested):
Hey @GuessWhoSamFoo thank you for the help on this! The site looks good, but the docs are pointing to a version that you're not using. Just a small change and we should be good to go.

PR #2388, opened 2021-04-30
Move managed fields in YAML tab to metadata

### LINKED ISSUE #2132: ManagedFields optional in YAML tab
Make managed fields optional in yaml output. Example: https://github.com/kubernetes/kubernetes/pull/96878/commits/237f051a5845275aaf7384aa6c4ec6ecce7614dd

GuessWhoSamFoo:
Since managed fields are immutable, it does not make sense to show them in the YAML tab which is used to edit objects especially since they take up a lot of space.

They are now moved under metadata:
image

- Fixes #2132 

A good video to understand purpose of managed fields: https://www.youtube.com/watch?v=P12z8X-qhis

Documentation: https://kubernetes.io/docs/reference/using-api/server-side-apply/

Is there a better way to represent JSON data with current components?


GuessWhoSamFoo:
We should use https://github.com/josdejong/jsoneditor to show managed fields.

internal/printer/metadata.go
```diff
@@ -118,3 +131,50 @@ func (m *Metadata) createSummary() (*component.Summary, error) {
...
 	return summary, nil
 }
+
+func (m *Metadata) managedFields() ([]component.Summary, error) {
+	a, err := meta.Accessor(m.object)
+	if err != nil {
+		return nil, err
+	}
+	if len(a.GetManagedFields()) == 0 {
+		return nil, nil
```
xtreme-vikram-yadav, line 141: We are not handling this case in the calling method which will fail to iterate over nil. This is omittable as not having managed fields will just return an empty array.

web/src/app/modules/shared/components/presentation/json-editor/json-editor.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  update() {
+    const view = this.v;
+    this.options.mode = view.config.mode;
+
+    try {
+      this.content = JSON.parse(view.config.content);
+    } catch (e) {
+      this.options.mode = 'preview';
+      this.content = {
+        error: 'cannot parse json',
```
xtreme-vikram-yadav, line 55: Not a huge issue, just wondering if this could be tested.

xtreme-vikram-yadav (changes requested):

web/src/app/modules/shared/components/presentation/json-editor/json-editor.component.ts
GuessWhoSamFoo: It is now tested

internal/printer/metadata.go
GuessWhoSamFoo: Removed

xtreme-vikram-yadav:
Sorry for bringing this up now but I find black font color taxing on eyes in dark mode and it looks a bit off compared to other text in dark mode. It would be nice if the font color could switch to light in dark mode mode and vice versa.

mklanjsek:
Nicely done, I like this approach!  However new component needs to be styled. In particular: border color, text font and color have to match the rest of the app. Dark theme is especially off with colors. As for font, ideally we should use same fixed font as in Monaco editor. 

The way double quotes are escaped is also not ideal, any chance that could be changed?

GuessWhoSamFoo:
Options that come to mind: raw string literals with backticks, string the output of decoded json byte slice, or fork go. What is a more developer friendly way to do this?

xtreme-vikram-yadav (approved):

PR #2423, opened 2021-05-07
Fix title component columns

GuessWhoSamFoo:
https://kubernetes.slack.com/archives/CM37M9FCG/p1620323334149200

### DIFF
--- a/web/src/app/modules/shared/components/presentation/tabs/tabs.component.html
+++ b/web/src/app/modules/shared/components/presentation/tabs/tabs.component.html
@@ -1,8 +1,8 @@
 <div class="clr-row clr-align-items-center clr-justify-content-between">
-  <div class="clr-col-6">
+  <div class="clr-col-7">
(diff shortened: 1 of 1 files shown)


mklanjsek (changes requested):
We can do better than that. The breadcrumbs should not break if no content is present on the right! I suggest setting the fixed `max-width` on button area and let the breadcrumbs take as much space as available.

GuessWhoSamFoo:
I don't see how `max-width` changes anything. Max width is already set with clarity grids. Breadcrumbs and title components can be arbitrarily wide, and they wrap into the columns if those widths are exceeded so breadcrumbs always remain visible.

mklanjsek:
Bottom line is that breadcrumbs should **never** break like this when there is so much empty space to the right:

Screen Shot 2021-05-07 at 1 20 17 PM

That just looks sloppy and confusing. There are many ways to fix it with css, does not have to be by setting the `max-width` - let me know if you need help with it.

PR #2425, opened 2021-05-07
Added SelectFile component

### LINKED ISSUE #2243: File Uploader - Drag and Drop or "Browse" filesystem
**Problem**
My team is building a commercial plugin for the [kpack](https://github.com/pivotal/kpack) project and would like to allow our users to upload a configuration file. Another use case @wwitzel3 mentioned would be for a user to upload a kubeconfig using a file picker instead of having to paste one in. 

**Proposed Solution**
image (9)
OR
download

@wwitzel3 mentioned we could potentially have a custom action registered that the file payload would get sent as part of, probably as a path to the location of the file on disk, so that we aren't sending the file over the wire (since octant is local). 

ref:tdc ref:tbs

mklanjsek:
Added a new SelectFile component intended to be used to select file(s) from the file system. 

Also added this new file selector to apply yaml, so it's now possible to select and open a yaml file before applying it.  While there, I refactored the form buttons to use `cds-button` and be consistent with the FileSelect button.

- Fixes #2243

### DIFF
--- a/pkg/view/component/select_file.go
+++ b/pkg/view/component/select_file.go
@@ -1,0 +1,68 @@
+package component
+
+import "encoding/json"
+
+type SelectFile struct {
+	Base
+	Config SelectFileConfig `json:"config"`
+}
+
+type Layout string
+
+const (
+	LayoutHorizontal Layout = "horizontal"
+	LayoutVertical   Layout = "vertical"
+	LayoutCompact    Layout = "compact"
+)
+
+type FileStatus string
+
+const (
+	FileStatusSuccess FileStatus = "success"
+	FileStatusError   FileStatus = "error"
+)
+
+type SelectFileConfig struct {
+	Label         string     `json:"label"`
+	Multiple      bool       `json:"multiple"`
+	Status        FileStatus `json:"status"`
+	StatusMessage string     `json:"statusMessage"`
+	Layout        Layout     `json:"layout"`
+	Action        string     `json:"action,omitempty"`
+}
+
+// NewSelectFile creates a new Select File component.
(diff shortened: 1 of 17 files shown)


pkg/view/component/select_file.go
```diff
@@ -0,0 +1,68 @@
+package component
+
+import "encoding/json"
```
GuessWhoSamFoo, line 3: We want to use `"github.com/vmware-tanzu/octant/internal/util/json"` here

lenriquez (approved):
Looks great to me!

web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
```diff
@@ -0,0 +1,73 @@
...
+  ViewChild,
+} from '@angular/core';
+import { AbstractViewComponent } from '../../abstract-view/abstract-view.component';
+import { SelectFileView } from '../../../models/content';
+import { WebsocketService } from '../../../../../data/services/websocket/websocket.service';
+import { ActionService } from '../../../services/action/action.service';
+
+@Component({
+  selector: 'app-view-select-file',
+  templateUrl: './select-file.component.html',
```
lenriquez, line 21: I wonder if it should switch the change detection to "ChangeDetectionStrategy.OnPush"

xtreme-vikram-yadav:
👍

web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
mklanjsek: I was considering it, but since it’s a simple component, I decided to keep it simple for now.

PR #2447, opened 2021-05-12
Improved Resource Viewer initial selection

### LINKED ISSUE #2408: Resource Viewer: initial selection of active resource
When Resource Viewer is opened for a particular resource, e.g. pod, that resource should be selected by default and info about that resource should be visible in details pane.  When Resource Viewer is displayed inside the Applications page, a first pod should be selected.

mklanjsek:
Resource Viewer now initially selects proper object depending on current view. This is  important for showing the right context and it will be even more important as we start adding more info to the details pane. Additionally, this change tweaks the graph padding, so if one or two boxes are present, they will not take the whole view.  

- Fixes #2408

### DIFF
--- a/internal/modules/workloads/detail_describer.go
+++ b/internal/modules/workloads/detail_describer.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"path"
+	"sort"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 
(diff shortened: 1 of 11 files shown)


xtreme-vikram-yadav (approved):
👍

internal/modules/workloads/detail_describer.go
```diff
@@ -124,7 +125,12 @@ _%s_
 		objects = append(objects, &pods[i])
 	}
 
-	rv, err := resourceviewer.Create(ctx, options.Dash, options.Queryer, objects...)
+	sort.Slice(objects, func(i, j int) bool {
+		return objects[i].GetName() < objects[j].GetName()
```
lenriquez, line 129: I don't really understand why sorting by name gives you the initial selection?

lenriquez (approved):
It looks great!

internal/modules/workloads/detail_describer.go
mklanjsek: It's important to sort pods, otherwise in consecutive updates selection will go from one pod to another, resulting in screen flicker.

PR #2478, opened 2021-05-20
show kube config path in preferences

### LINKED ISSUE #1893: Kubeconfig location on windows
Currently when octant on widows and starting without a kubeconfig, it pops up with the text area to paste a kubeconfig content. this content is then saved in the users AppData\Local\Temp\1 folder as a file named kubeconfig and then a number sequence.
the issue is the location of that file is not trivial for many users and i haven't seen it documented so updating that kubeconfig becomes difficult. also as that kubeconfig is loaded again at every start up of octant it becomes very difficult and problematic for users when they provision a new cluster etc.

there are a few possible options isee:
1. this is also in my opinion the best option from my side - add a menu option to edit that kubeconfig or paste a new kubeconfig content from a running instance of octant. basically allow a user to request the enter kubeconfig form to pop up on demand.
2. Save the file in a easier to find location C:\Users\<USERNAME>\.octant\kubeconfig maybe?
3. document where the file is saved so at least it would be easier to find.

This issue seems to be only really relevant for windows however if solution 1 was implemented i could see it being beneficial cross OS.

**Environment:**

- Octant version (use `octant version`): nightly electron build for windows
- Kubernetes version (use `kubectl version`): 1.17-1.20
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): Windows 10

xtreme-vikram-yadav:
**What this PR does / why we need it**: This PR assists the user in finding the local path to kube config which is used by octant. It is useful is case when the user has uploaded the config manually via loading api and a temporary kube config file is created and stored in the temp directory.

- Fixes #1893 

Steps for reproduction/acceptance:
```
Given kube config exists in a default location(~/.kube/config)
When the user starts octant
And the user opens preferences
Then the user sees the local kube config path with icon to copy to clipboard
```
```
Given kube config does not exist in a default location(~/.kube/config)
When the user starts octant
And the user pastes the kube config in the modal
And the user opens preferences
Then the user sees the local kube config path with icon to copy to clipboard
```

Some implementation decisions:

The `octant.generator` interface now returns multiple events which extends a *_manager's capability to perform multiple small computations and stream back multiple results as events. A managers setup is an overkill for implementing small features. This approach helps with testing by keeping the handlerFunc usage the same as before and avoids generator classes doing duplicated work.

### DIFF
--- a/changelogs/unreleased/2478-xtreme-vikram-yadav
+++ b/changelogs/unreleased/2478-xtreme-vikram-yadav
@@ -1,0 +1,1 @@
+Updated preferences to show local path to kube config
--- a/internal/api/context_manager.go
+++ b/internal/api/context_manager.go
@@ -25,7 +25,7 @@ import (
 type ContextManagerOption func(manager *ContextManager)
 
 // ContextGenerateFunc is a function which generates a context event.
-type ContextGenerateFunc func(ctx context.Context, state octant.State) (oevent.Event, error)
+type ContextGenerateFunc func(ctx context.Context, state octant.State) ([]oevent.Event, error)
 
 // WithContextGenerator sets the context generator.
 func WithContextGenerator(fn ContextGenerateFunc) ContextManagerOption {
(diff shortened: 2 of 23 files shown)


mklanjsek:
I am not sure if this is the bast way to solve the problem from user experience point of view. Maybe just making `kube.config` location directly configurable from preferences would be better and in the same time it would allow users to switch between different configurations. We should probably discuss it as a team.

xtreme-vikram-yadav:
I created https://github.com/vmware-tanzu/octant/issues/2497 to allow users to upload their kube config.

mklanjsek:
Looks good! You'll need to regenerate mock files before merging.

xtreme-vikram-yadav:
I think it's an environment thing where there are differences in the `go generate` output. I'll use a linux machine this time to fix it and think of a long term solution to notify of the version differences for tools like `mockgen` and `goimports` in the CI.

mklanjsek (approved):

PR #2498, opened 2021-05-26
Fix bug where actions are not passed to components

### LINKED ISSUE #2458: Card actions are not firing
Cards have to ability to embed forms and use a component action. These currently broke with the form changes.

Maybe we shouldn't use this at all since such patterns have not been seen outside of Octant internals, and it is not discoverable to users.

**Environment:**

- Octant version (use `octant version`): 0.20

GuessWhoSamFoo:
Non-numbered action form fields were filtered by the transform function.

- Fixes #2458 

### DIFF
--- a/web/src/app/modules/shared/models/form-helper.spec.ts
+++ b/web/src/app/modules/shared/models/form-helper.spec.ts
@@ -1,0 +1,69 @@
+import { FormHelper } from './form-helper';
+
+describe('FormHelper', () => {
+  let formHelper: FormHelper;
+
+  beforeEach(() => {
+    formHelper = new FormHelper();
+  });
+
+  it('converts number', () => {
+    expect(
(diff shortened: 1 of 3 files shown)


web/src/app/modules/shared/models/form-helper.spec.ts
```diff
@@ -0,0 +1,39 @@
+import { FormHelper } from './form-helper';
+
+describe('FormHelper', () => {
```
xtreme-vikram-yadav, line 3: Since we are testing return value, we can easily add tests for values with `empty string`, `stringified number` and `NaN`.

GuessWhoSamFoo: done

wwitzel3 (approved):

PR #2517, opened 2021-06-01
Rename Expandable Row Detail TS Component Config Variable

liamrathke:

This PR is part of a larger update to the TypeScript Plugin library. While attempting to update the TypeScript components available in the library, a specific component (`ExpandableRowDetail`) failed to build, specifically because its configuration name did not match the format that the generator expected (it was looking for `Config ExpandableRowDetailConfig` when the config was actually just named `ExpandableDetailConfig`). Merging this PR will allow us to update the TypeScript component library such that it works correctly in new TypeScript-based plugins going forward. 

[Relevant comment](https://github.com/vmware-tanzu/plugin-library-for-octant/pull/21#issuecomment-852148257
) in `vmware-tanzu/plugin-library-for-octant` by @GuessWhoSamFoo

- Fixes https://github.com/vmware-tanzu/plugin-library-for-octant/issues/20

I investigated adding some additional error handling to better inform developers when these naming conflicts emerge, but realized that this could be somewhat complicated because the template generating process involves running commands, so the error message might need to be passed through a standard input and then parsed. 

**Release note**:
```
release-note
```

☝️ not sure if this section needs to be filled out, please let me know if there's info I should put here

### DIFF
--- a/pkg/view/component/row_detail.go
+++ b/pkg/view/component/row_detail.go
@@ -23,7 +23,7 @@ type ExpandableRowDetail struct {
 	Config ExpandableRowDetailConfig `json:"config"`
 }
 
-// ExpandableDetailConfig is a configuration for an expandable row detail.
+// ExpandableRowDetailConfig is a configuration for an expandable row detail.
 type ExpandableRowDetailConfig struct {
 	Replace bool        `json:"replace"`
 	Body    []Component `json:"body"`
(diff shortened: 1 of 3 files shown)


liamrathke:
Also, not sure what is going on with the DCO, I followed the steps the workflow told me to take to resolve the issue, but somehow it's still failing, is there a way to re-trigger the GitHub checks?

GuessWhoSamFoo (approved):

PR #2519, opened 2021-06-02
Make properties pane resizable

mklanjsek:
Resource Viewer properties sometimes contain long fields that are not easy to read. To address that, Properties pane is now resizable. By default, it takes  25% of parent width and can be resized to contain anything from 20% - 70%.

https://user-images.githubusercontent.com/34245594/120484547-6d27a400-c370-11eb-8a53-bfb57d3dfaa7.mov

Also tweaked css parameters to address minor cosmetic issues.  

### DIFF
--- a/web/src/app/modules/shared/components/presentation/resource-viewer/resource-viewer.component.ts
+++ b/web/src/app/modules/shared/components/presentation/resource-viewer/resource-viewer.component.ts
@@ -5,9 +5,12 @@
 import {
   ChangeDetectorRef,
   Component,
+  ElementRef,
   isDevMode,
   OnDestroy,
   OnInit,
+  Renderer2,
+  ViewChild,
   ViewEncapsulation,
 } from '@angular/core';
 import {
@@ -20,6 +23,7 @@ import { ELEMENTS_STYLE, ELEMENTS_STYLE_DARK } from './octant.style';
 import { Router } from '@angular/router';
 import { ThemeService } from '../../../services/theme/theme.service';
 import { Subscription } from 'rxjs';
+import { ResizeEvent } from 'angular-resizable-element';
 
 const statusColorCodes = {
   ok: '#60b515',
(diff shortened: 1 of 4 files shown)


web/src/app/modules/shared/components/presentation/resource-viewer/resource-viewer.component.ts
```diff
@@ -173,4 +180,36 @@ export class ResourceViewerComponent
...
+      topOrBottom: 'ns-resize',
+    };
+  }
+
+  resizeEnd() {
+    this.zoom = Object.assign({}, defaultZoom); // update without layout recalc
+  }
+
+  resizeStart() {
+    this.startPosition = this.elRef.nativeElement.firstChild.firstChild.offsetWidth;
```
xtreme-vikram-yadav, line 200: Accessing children this way is a bit hard to read. A couple of options to make it readable would be to:
1) Assign these references to variables based on the element identifiers (`viewContainer` in this case).
2) Define ViewChild 
```
 @ViewChild('viewContainer')
  viewContainer: ElementRef;

  @ViewChild('resourceViewer')
  resourceViewer: ElementRef;

  @ViewChild('statusContainer')
  statusContainer: ElementRef;
```
and then use renderer (Renderer2) to update the style (`this.renderer.setStyle(this.viewContainer.nativeElement, 'width', `${leftSize}%`)`);

xtreme-vikram-yadav (changes requested):

web/src/app/modules/shared/components/presentation/resource-viewer/resource-viewer.component.ts
mklanjsek: Good point, Vikram - should have thought of that!

mklanjsek:
@xtreme-vikram-yadav PTAL

xtreme-vikram-yadav (approved):

Issue #2383, opened 2021-04-29
Ability to inject custom objectstore into an octant runner


pivotal-mikegresham:
**Describe the problem/challenge you have**

We have a bunch kubernetes object data, passively collected in a database by an aggregation system, and exposed via an authenticated CRUD API. We would like to visualize/interact with this data via Octant + plugins. To do this, We are consuming Octant's `dash.NewRunner` function as a library function, and running its streaming websocket API.

**Describe the solution you'd like**

We'd like to be able to pass an alternate `store.Store` implementation (containing a client to the abovementioned CRUD API) to `NewRunner` for our octant server to use while dispatching `dashboardClient` requests from plugins.

**Anything else you would like to add:**

Given the precedent set by the [`WithClusterClient` functional option](https://github.com/vmware-tanzu/octant/blob/a73ca4222088129113e16c6815b70a72c80e58be/pkg/dash/dash.go#L210-L216), a similar `WithObjectStore` option seems fair.

Something we've encountered while experimenting with implementing the `store.Store` interface is that the signature of the [`UpdateClusterClient` method](https://github.com/vmware-tanzu/octant/blob/a73f0e06766be2450a30780ae3ec3f1272ca83da/pkg/store/store.go#L41) mentions the `ClientInterface` type from the `internal/cluster` package. So for external consumers of Octant to provide their own Store, either

* this type reference would need to be modified, or
* the `ClientInterface` type would need to be moved into a non-internal package

This is currently blocking us from using upstream Octant, so we'd like to see this implemented in 0.21 or as soon as possible.

**Environment:**

- Octant version (use `octant version`): 0.19
- Kubernetes version (use `kubectl version`): 1.19+
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): k8s container on AWS

ref: tdc

PR #2520, opened 2021-06-02
Add ObjectStore functional option.

wwitzel3:

- Export the cluster.ClientInterface so that it can be properly injected at runtime with functional options `WithClusterClient` previously depended on an internal interface requiring a fork of Octant to use.
- Add a functional options to inject an ObjectStore at runtime, the store.Store interface was already an external interface.

Depends on #2581


jamieklassen:
This fixes #2383

wwitzel3:
moved #2610

PR #2522, opened 2021-06-02
Refactor dyamic cache to use namespaced informers

### LINKED ISSUE #1172: Get blank contents in overview page and repeating exception in error log.

Am running Octant inside of a container for training purposes. It has a restricted RBAC. Neither of these appear to be the issue, as mostly it works, but in certain cases it fails when initially showing the Overview page for the namespace.

From various tests it seems that if Octant is started against an empty namespace (just default service account, token secret, some role bindings), it works fine from that point on. If however you deploy a workload before accessing Octant web interface (although Octant was started before the workloads were created), then it fails to display the overview page.

The URL directly targeted when accessing Octant the first time is of the form:

```
https://xxx/#/overview/namespace/lab-octant-testing-w01-s015
```

An explicit URL to the namespace was used when accessing it the first time, as on older versions if you used the root of the site it was being hit by rate limiting as it would query all namespaces and the contents of them, which would fail because of the restricted RBAC.

Result in the web interface is:

image

The error logs are filled with:

```
2020-07-28T06:43:22.732Z        ERROR   api/content_manager.go:130      generate content        {"client-id": "9fbca94d-d09d-11ea-825e-0242ac110009", "err": "generate content: scheme convert: Object 'Kind' is missing in 'unstructured object has no kind'", "content-path": "overview/namespace/lab-octant-testing-w01-s015"}
        /__w/octant/octant/internal/api/content_manager.go:130
        /__w/octant/octant/internal/api/poller.go:86
        /__w/octant/octant/internal/api/poller.go:95
        /__w/octant/octant/internal/api/content_manager.go:106
```

Tried switching back to connecting the first time by using root URL for site to see if makes a difference, but it doesn't and still fails in the way described.

For the purposes of the training/demo, as a workaround, means have to tailor content to ensure people access Octant before deploying a workload. If this is done, and the workload only deployed after Octant is accessed the first time, then it works okay.

Expect the overview for the namespace to be shown.

Am using ``--namespace`` option to have it start up against specific namespace as well.

**Environment:**

```
[~/exercises] $ octant version
Version:  0.14.1
Git commit:  51ce7f6a7e87c9cfcabc9bf402dff7eda5d6986b
Built:  2020-07-20T23:52:37Z
[~/exercises] $ kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.9", GitCommit:"4fb7ed12476d57b8437ada90b4f93b17ffaeed99", GitTreeState:"clean", BuildDate:"2020-07-15T16:18:16Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:35:47Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
[~/exercises] $ uname -a
Linux lab-octant-testing-w01-s017-65f9bb66cf-s7qmk 4.19.107 #1 SMP Thu Mar 26 11:33:10 PDT 2020 x86_64 x86_64 x86_64 GNU/Linux
```

GuessWhoSamFoo:
Adds more clarity around the nuances of the informer cache and why various informer factories were used throughout the project. Also cleans up misc console warnings/errors.

- Fixes #1172 

Binary is available `curl -Lo octant https://storage.googleapis.com/octant-nightlies/octant` and can run in the RBAC restricted lab environment via `./octant --namespace $SESSION_NAMESPACE --disable-open-browser --listener-addr 127.0.0.1:10086 --accepted-hosts $SESSION_NAMESPACE-console.$INGRESS_DOMAIN`


wwitzel3:
I would like to see the tracing output of this running against our development cluster, before and after.

GuessWhoSamFoo:
Not a scientific approach, it looks like the times to List on overview pages are roughly the same. For reference, this is on the team's development cluster on the default namespace.

0.20.0:
image

This branch:
image

Informers for CRD's are not shutting down after `TestNewRunnerUsesClusterClient` and causing a flake: `E0602 14:39:15.354830 3784560 runtime.go:76] Observed a panic: Fail in goroutine after TestNewRunnerUsesClusterClient has completed`. Taking a look into the test

internal/objectstore/dynamic_cache.go
```diff
@@ -453,33 +419,16 @@ func (d *DynamicCache) gvrFromKey(ctx context.Context, key store.Key) (schema.Gr
...
-
-		// do not store an empty GVR
-		if gvr.Version == "" && gvr.Resource == "" {
-			return schema.GroupVersionResource{}, fmt.Errorf("unable to locate GVR for GVK: %s", key.GroupVersionKind())
-		}
-
-		d.gvrCache.Store(gk, gvr)
-	} else {
-		gvr, _ = v.(schema.GroupVersionResource)
+	gvr, _, err := d.client.Resource(gk)
```
wwitzel3, line 429: How do we feel about creating a static list, like we have for GVKs that makes the Key to a GVR for known types and attempts to look it up for unknown types?

internal/printer/persistentvolumeclaim.go
```diff
@@ -260,7 +260,8 @@ func getBoundPersistentVolume(ctx context.Context, pvc *corev1.PersistentVolumeC
 
 	pv, err := objectStore.Get(ctx, key)
 	if err != nil {
-		return nil, errors.Wrapf(err, "get volume for key %+v", key)
+		// Return empty persistent volume instead of error in case the dynamic cache has not found the volume yet
+		return persistentVolume, nil
```
wwitzel3, line 264: I remember doing this way long ago to make the educates usage of Octant work, that we have bring this back in to make it work is a sign that things still aren't quite right.

We should be handling these errors in a better way.

Imagine using Octant thinking you had access to PVs or knowing that you created some but you just see empty .. no error, no indication that there was an error of anykind.

pkg/dash/dash.go
```diff
@@ -394,8 +394,8 @@ func (r *Runner) initAPI(ctx context.Context, logger log.Logger, opts ...RunnerO
 
 	logger.Debugf("initial namespace for dashboard is %s", options.Namespace)
 
-	factoryOption := objectstore.WithDynamicSharedInformerFactory(options.factory)
-	appObjectStore, err := initObjectStore(ctx, clusterClient, factoryOption)
+	// factoryOption := objectstore.WithDynamicSharedInformerFactory(options.factory)
```
wwitzel3, line 397: delete

pkg/dash/dash_test.go
```diff
@@ -173,60 +169,6 @@ func TestNewRunnerLoadsValidKubeConfigFilteringNonexistent(t *testing.T) {
 	require.Equal(t, "test-context", kubeConfigEvent.Data["currentContext"].(string))
 }
 
-func TestNewRunnerUsesClusterClient(t *testing.T) {
```
wwitzel3, line 176: Why is these test being deleted? That wasn't clear to me?
If the watcher for CRDs are not shutting down properly I think we want to fix what ever is causing that.

internal/objectstore/dynamic_cache.go
GuessWhoSamFoo: Turns out the special case for CRDs was due to improperly passing the version to RESTMapper.

We probably don't need this except for testing via https://github.com/kubernetes/client-go/blob/master/dynamic/fake/simple.go#L43

internal/printer/persistentvolumeclaim.go
GuessWhoSamFoo: This behavior is now changed to show the volume as text

image

pkg/dash/dash_test.go
GuessWhoSamFoo: Added back this test

wwitzel3 (approved):

PR #2540, opened 2021-06-07
Add flag to disable cross origin check

### LINKED ISSUE #2384: Flexible origin-checking in octant for running behind a reverse proxy

We are running a hosted Octant websocket server behind a reverse proxy which routes based on hostname - this means that, by the time requests from a browser reach Octant, the Host has been rewritten by the proxy and no longer matches the Origin.

Some way to inject a custom origin-checking function, to substitute for the [hardcoded `checkSameOrigin`](https://github.com/vmware-tanzu/octant/blob/994b705a04f68fbbf6c56aaf0da91bfd0c512b54/internal/api/middleware.go#L48-L79). Ideally this would take the form of a functional option to `dash.NewRunner` as that is the entrypoint we are using to run our Octant server.

the `Upgrader` type from `gorilla/websocket` has a field for a [`CheckOrigin` function](https://github.com/gorilla/websocket/blob/c3dd95aea9779669bb3daafbd84ee0530c8ce1c1/server.go#L59-L66) -- this might be a good interface to follow. Actually it looks like Octant doesn't currently make use of that field but maybe it could, and allow it to be overridden via a functional option.

This is one of the thing blocking us from getting back to upstream Octant from our fork, so we'd like to see it in 0.21 or as soon as possible.

**Environment:**

- Octant version (use `octant version`): 0.19
- Kubernetes version (use `kubectl version`): 1.19+
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc): k8s container on AWS

ref: tdc

GuessWhoSamFoo:
Adds `--disable-origin-check` flag or `OCTANT_DISABLE_ORIGIN_CHECK` env var to skip checks in `rebindHandler`.

This is done instead of adding a configuration upgrader function in dash.go because we want to move away from using it as a catch-all.

cc @jamieklassen Let me know if this works for your use case

- Fixes #2384 

### DIFF
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -116,11 +116,11 @@ func rebindHandler(ctx context.Context, acceptedHosts []string) mux.MiddlewareFu
 			if disableCheckOrigin := viper.GetBool("disable-origin-check"); !disableCheckOrigin && !checkSameOrigin(r) {
 				logger := log.From(ctx)
 				logger.Debugf("check same origin failed")
-				httpErrors = append(httpErrors, "forbidden bag origin")
+				httpErrors = append(httpErrors, "forbidden bad origin")
 			}
 
 			if len(httpErrors) > 0 {
-				http.Error(w, strings.Join(httpErrors, ":"), http.StatusForbidden)
+				http.Error(w, strings.Join(httpErrors, ": "), http.StatusForbidden)
 				return
 			}
 
--- a/internal/api/middleware.go
+++ b/internal/api/middleware.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"unicode/utf8"
 
+	"github.com/spf13/viper"
+
 	"github.com/gorilla/mux"
 
 	"github.com/vmware-tanzu/octant/internal/log"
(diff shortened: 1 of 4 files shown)


internal/api/middleware.go
```diff
@@ -104,17 +106,21 @@ func rebindHandler(ctx context.Context, acceptedHosts []string) mux.MiddlewareFu
...
-				return
+				httpErrors = append(httpErrors, "forbidden host")
 			}
 
-			if !checkSameOrigin(r) {
+			if disableCheckOrigin := viper.GetBool("disable-origin-check"); !disableCheckOrigin && !checkSameOrigin(r) {
 				logger := log.From(ctx)
 				logger.Debugf("check same origin failed")
-				http.Error(w, "forbidden bad origin", http.StatusForbidden)
+				httpErrors = append(httpErrors, "forbidden bag origin")
```
jamieklassen, line 119: typo
```suggestion
				httpErrors = append(httpErrors, "forbidden bad origin")
```

internal/api/middleware.go
```diff
@@ -104,17 +106,21 @@ func rebindHandler(ctx context.Context, acceptedHosts []string) mux.MiddlewareFu
...
+			}
+
+			if len(httpErrors) > 0 {
+				http.Error(w, strings.Join(httpErrors, ":"), http.StatusForbidden)
```
jamieklassen, line 123: just my two cents, but it seems like the spirit of this is to give requests feedback about all the possible reasons they might be rejected, and the specific approach allows for responses like `forbidden host:forbidden bad origin`. Maybe merits a unit test?

jamieklassen (commented):
This definitely covers our functional requirement -- if we deploy using this flag we'll be able to avoid getting 403ed all the time, but we won't be able to genuinely verify request origins. It's not the biggest deal because we've been using a patched version of octant with a similar workaround already.

@GuessWhoSamFoo when you say

> move away from using it as a catch-all

do you mean using dash.go as a catch-all? or using an injected upgrader as a catch-all? What's the significance of "catch-all" here, like using a feature for unintended use cases?

GuessWhoSamFoo:
The other way this could've been written is through a wrapper around upgrader in `websocket_service.go`. It would end up with `WithUpgrader(func(r *http.Request) RunnerOption {...}` and then passing into `initAPI` though this `dashConfig` struct that's passed around to almost all of `internal/`.

The difference is not being able to enable the flag using the `With...` pattern seen so far. It's seems to be mostly a matter of preference so happy to shape into whatever is easier.

jamieklassen:
I feel like there are two questions afoot here:
1. in what way should origin checking be configurable -- should library consumers be able to provide an arbitrary websocket upgrader, or should there only be a flag for turning the origin check on or off?
2. how should this configuration option be managed -- via a viper config or a Runner field?

(this might be obvious, but I think a pointer type like an arbitrary websocket upgrader cannot be stored in a viper config, so that combination of solutions is invalid)

I think your first message where you talked about "catch-alls" was a comment about 1. above, whereas your more recent message relates to 2. Am I understanding correctly? Are you looking for input from me about this?

GuessWhoSamFoo:
Cool, I think we're on the same page - thanks for taking a look! We'll get this in 0.21

mklanjsek (approved):

PR #2546, opened 2021-06-10
Return file metadata on file upload

GuessWhoSamFoo:
Creates a type interface for files to send metadata and added types for typescript.

Also fixes build.go to work on a clean build environment because the ordering of commands were incorrect.

### DIFF
--- a/web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
+++ b/web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
@@ -13,8 +13,16 @@ import {
 } from '@angular/core';
 import { AbstractViewComponent } from '../../abstract-view/abstract-view.component';
 import { SelectFileView } from '../../../models/content';
-import { WebsocketService } from '../../../../../data/services/websocket/websocket.service';
 import { ActionService } from '../../../services/action/action.service';
+import { ElectronService } from '../../../services/electron/electron.service';
+
+type File = {
+  name: string;
+  type: string;
+  path: string;
+  lastModified: number;
+  size: number;
+};
 
 @Component({
   selector: 'app-view-select-file',
@@ -33,7 +41,10 @@ export class SelectFileComponent
   @ViewChild('fileInput') fileInput: ElementRef;
   @Output() fileChanged: EventEmitter<any> = new EventEmitter<any>();
 
-  constructor(private actionService: ActionService) {
+  constructor(
+    private actionService: ActionService,
+    private electronService: ElectronService
+  ) {
     super();
   }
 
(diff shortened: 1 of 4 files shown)


web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
```diff
@@ -47,19 +58,31 @@ export class SelectFileComponent
...
+          lastModified: file.lastModified,
+          size: file.size,
+        });
       }
 
       if (this.action) {
         this.actionService.perform({
           action: this.action,
-          files: event.target.files,
+          files: fileList,
```
mklanjsek, line 85: The intention of the original implementation was to pass the selection data as is, and allow consumer to directly access not only the selected file(s) metadata (defined in `lib.dom.d.ts`), but also the file content thought `Blob` interfaces. This approach makes copy of original file metadata, and removes ability to read file with original `Blob` interfaces. This is considerable loss for typescript consumer - how are they going to read the file with our metadata that does not contain either path info or `Blob` steaming interfaces?

web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
```diff
@@ -13,8 +13,16 @@ import {
...
 import { AbstractViewComponent } from '../../abstract-view/abstract-view.component';
 import { SelectFileView } from '../../../models/content';
-import { WebsocketService } from '../../../../../data/services/websocket/websocket.service';
 import { ActionService } from '../../../services/action/action.service';
+import { ElectronService } from '../../../services/electron/electron.service';
+
+type File = {
+  name: string;
+  type: string;
+  path: string;
```
mklanjsek, line 22: `path` is not passed from browser for security reason. This limits action consumer ability to read the file (see below)

web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
```diff
@@ -13,8 +13,16 @@ import {
 } from '@angular/core';
 import { AbstractViewComponent } from '../../abstract-view/abstract-view.component';
 import { SelectFileView } from '../../../models/content';
-import { WebsocketService } from '../../../../../data/services/websocket/websocket.service';
 import { ActionService } from '../../../services/action/action.service';
+import { ElectronService } from '../../../services/electron/electron.service';
```
mklanjsek, line 17: not used anywhere?

mklanjsek (changes requested):

web/src/app/modules/shared/components/presentation/select-file/select-file.component.ts
GuessWhoSamFoo: The file object isn't serializable. See: https://stackblitz.com/edit/angular-ivy-85wt7e?file=src/app/app.component.ts


Under existing usage, we  use a different code path entirely (https://github.com/vmware-tanzu/octant/blob/master/web/src/app/modules/shared/components/smart/editor/editor.component.ts#L92). Consumers of this action will have to take the filepath  from this action and read it locally . In other words, this component would only see meaningful usage on electron.


iirc, it was decided explicitly not to send the file over the wire.

mklanjsek (approved):
I would prefer if we have a solution that works better for typescript plugin authors, but let's go with this.

GuessWhoSamFoo:
Let's make the decision-making explicit: is the intention of this component to send a file over the wire?

mklanjsek:
No, but if a consumer wants to download a file it would be good to provide browser native streaming interfaces, of course when that's possible.
________________________________

GuessWhoSamFoo:
There are no instances of `upload` being used in documentation or the code base. This component (with exception of applying yaml as that uses a different code path) will only be used for getting a local file path unless requested otherwise.

Issue #2547, opened 2021-06-10
New errors at Octant startup and when opening custom resources


mklanjsek:
After upgrading my codebase to current master (70d7a5a0ad64a273e380cb76116526cd4c3a2403), I am seeing some new errors popping up:

1. At the startup, this error is repeated multiple times, not sure if it has any implications though:

```
2021-06-10T10:42:13.420-0600	WARN	objectstore/dynamic_cache.go:415	unable to start watcher %!(EXTRA string=failed to list *unstructured.Unstructured: conversion webhook for messaging.knative.dev/v1, Kind=InMemoryChannel failed: Post "https://webhook.mink-system.svc:443/resource-conversion?timeout=30s": x509: certificate signed by unknown authority)	{"component": "websocket-client", "client-id": "cbb4d5ba-ca0a-11eb-8323-acde48001122", "poller-name": "navigation", "poller-instance": "d40801d8-4fbe-4331-91ab-d3fd927fc3e5"}
github.com/vmware-tanzu/octant/internal/objectstore.(*DynamicCache).watchErrorHandler.func1
	/Users/mklanjsek/workspace/octant/internal/objectstore/dynamic_cache.go:415
k8s.io/client-go/tools/cache.(*Reflector).Run.func1
	/Users/mklanjsek/workspace/octant/vendor/k8s.io/client-go/tools/cache/reflector.go:210
k8s.io/apimachinery/pkg/util/wait.BackoffUntil.func1
	/Users/mklanjsek/workspace/octant/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:155
k8s.io/apimachinery/pkg/util/wait.BackoffUntil
	/Users/mklanjsek/workspace/octant/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:156
k8s.io/client-go/tools/cache.(*Reflector).Run
	/Users/mklanjsek/workspace/octant/vendor/k8s.io/client-go/tools/cache/reflector.go:208
k8s.io/apimachinery/pkg/util/wait.(*Group).StartWithChannel.func1
	/Users/mklanjsek/workspace/octant/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:56
k8s.io/apimachinery/pkg/util/wait.(*Group).Start.func1
	/Users/mklanjsek/workspace/octant/vendor/k8s.io/apimachinery/pkg/util/wait/wait.go:73
```

[...]

PR #2564, opened 2021-06-16
Revert "Refactor dyamic cache to use namespaced informers"

GuessWhoSamFoo:
Reverts vmware-tanzu/octant#2522


mklanjsek (approved):
Verified that the CRD problem reported in #2547 does not happen on this branch ✅

PR #2582, opened 2021-06-23
Show default radio selection

GuessWhoSamFoo:
A form with multiple field input choices for radio buttons currently never reaches the code path for checking if it is checked already.

### DIFF
--- a/web/src/app/modules/shared/components/smart/editor/editor.component.ts
+++ b/web/src/app/modules/shared/components/smart/editor/editor.component.ts
@@ -137,6 +137,6 @@ export class EditorComponent
   }
 
   ngOnDestroy() {
-    this.subscriptionTheme.unsubscribe();
+    this.subscriptionTheme?.unsubscribe();
   }
 }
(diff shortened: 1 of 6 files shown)


web/src/app/modules/shared/components/smart/editor/editor.component.ts
```diff
@@ -137,6 +137,8 @@ export class EditorComponent
   }
 
   ngOnDestroy() {
-    this.subscriptionTheme.unsubscribe();
+    if (this.subscriptionTheme) {
+      this.subscriptionTheme.unsubscribe();
```
mklanjsek, line 141: Not a biggie, but `this.subscriptionTheme.unsubscribe();` would be more elegant and easier to read.

mklanjsek (approved):

web/src/app/modules/shared/components/smart/editor/editor.component.ts
GuessWhoSamFoo: The test in karma is not guaranteed to have subscribed to the service on tear down.

mklanjsek: Sorry, that was supposed to be `this.subscriptionTheme?.unsubscribe();` 😊

PR #2583, opened 2021-06-23
Adds form component to generated components

liamrathke:

The TypeScript component generation code was failing in part because the `Form` component was not marked as generatable. This adds necessary supplementary classes and definitions for the `Form` component to be generated.

**Testing steps**:

1. Checkout the fork
2. Inside `octant`, run `go run ./cmd/ts-component-gen/main.go -dest <directory>`, where `<directory>` is a path to a valid directory; verify that the command completes without issues/errors

**Additional notes**:
After merging, the generator command will need to be rerun inside `vmware-tanzu/plugin-library-for-octant`. If the [cron job PR](https://github.com/vmware-tanzu/plugin-library-for-octant/pull/24) is merged before then, no manual action will be needed.


GuessWhoSamFoo:
It looks like the ts generator does not do well with nested structs. Instead of changing the schema of forms again, FormFieldConfig should be changed generically to `Component` and check the conversion during unmarshaling.

PR #2586, opened 2021-06-25
Add ADOPTERS file

davidmirror-ops:
Signed-off-by: davidmirror-ops <daespejo@vmware.com>

**What this PR does / why we need it**: This PR will add an ADOPTERS file to showcase Octant adopters

- Fixes #

**Release note**:
```
release-note
```


ADOPTERS.md
```diff
@@ -0,0 +1,14 @@
+# Octant Adopters
+
+## Success Stories
+
+Below is a list of adopters of Octant in **production environments** that have
```
wwitzel3, line 5: Octant's target and focus is at developers. That said, you don't really "deploy" Octant so "in" seems like an odd term to use.

ADOPTERS.md
```diff
@@ -0,0 +1,14 @@
...
+publicly shared the details of how they use it.
+
+## Solutions built with Octant
```
wwitzel3, line 8: Solutions built with Octant would be plugins? .. This language is a bit confusing to me. There will probably never be very many solutions where Octant is being used at a component .. maybe some of the libraries we spin out of Octant eventually would fit that bill.

ADOPTERS.md
```diff
@@ -0,0 +1,14 @@
+# Octant Adopters
```
wwitzel3, line 1: Adopters is a bit odd, mainly because you don't really adopt it .. this is like saying 'adopters of kubectl' ...

wwitzel3 (changes requested):

ADOPTERS.md
jonasrosland: It's meant to be seen as "users". We have the same terminology in other projects: https://github.com/vmware-tanzu/velero/blob/main/ADOPTERS.md

ADOPTERS.md
jonasrosland: I think plugins make sense here, for sure

ADOPTERS.md
jonasrosland: We do want to know if people do use Octant to verify/troubleshoot etc their production environments, right?

davidmirror-ops:
Latest commit is a revised and condensed version of the ADOPTERS file text. If it's approved, it will be used as the baseline to start adding content to it.

ADOPTERS.md
```diff
@@ -0,0 +1,15 @@
...
+
+## Success Stories
+
+Below is a list of organizations that are using Octant to gain insight on their Kubernetes deployments and have publicly shared the details of how they use it.
+
+
+## Solutions built with Octant
+
+Below is a list of plugins created to extend the Octant feature set for additional use cases.
```
wwitzel3, line 10: We also have a place to list plugins over in our `reference.octant.dev` website and I believe we talked about listing plugins under the `Community` section of the main `octant.dev` website. Is there a plan to reduce these down to a single location?

Is this document going to be the location? Or will this just link out to one of those other places?

wwitzel3 (approved):
I had a question, but it isn't a blocker.

PR #2587, opened 2021-06-27
errors - explictly check for AccessErrors for CRDs and Events

wwitzel3:

Octant is currently broken with respect to partial resource loading on overview and summary pages.

Currently when attempting to load overview and summary pages if any single item in the listing fails, the entire content response is replaced with an error page. In the case of Overview pages, an emptyContentResponse is used and the page is stuck with the loading icon, even though some of the state was able to be fetched.

This fix is not ideal and a bit of shotgun anti-pattern, but it also unbreaks a lot of users in the short term and it important for restoring the functionality of Octant


wwitzel3:
Tested it some more and it isn't quite ready yet.

lenriquez (approved):
LGTM

PR #2610, opened 2021-06-30
Export config.Dash and related interfaces

### LINKED ISSUE #2383: Ability to inject custom objectstore into an octant runner

wwitzel3:

Replaces #2520 
Fixes #2383


pkg/plugin/api/client.go
```diff
@@ -15,14 +15,15 @@ import (
...
 	"github.com/vmware-tanzu/octant/pkg/action"
-	"github.com/vmware-tanzu/octant/pkg/config"
 	"github.com/vmware-tanzu/octant/pkg/plugin/api/proto"
 	"github.com/vmware-tanzu/octant/pkg/store"
 )
 
 //go:generate mockgen -destination=./fake/mock_dashboard_client.go -package=fake github.com/vmware-tanzu/octant/pkg/plugin/api/proto DashboardClient
 //go:generate mockgen -destination=./fake/mock_dashboard_connection.go -package=fake github.com/vmware-tanzu/octant/pkg/plugin/api DashboardConnection
 
+const MaxMessageSize int = 1024 * 1024 * 16
```
GuessWhoSamFoo, line 25: Can `pkg/config/config.go` be removed since this is done to avoid an import cycle and import `"github.com/vmware-tanzu/octant/pkg/plugin/api"` out of `commands/dash.go`?

GuessWhoSamFoo (approved):
Probably can phrase the change log entry better to avoid editing later

PR #2646, opened 2021-07-13
Bump to storybook 6.3.4 and compodoc 1.1.13

GuessWhoSamFoo:
The compodoc bump removes the need for `marked` to exist in resolutions although Github stopped warning about it.

Fixes deprecated import in `select.file.stores.mdx` https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#deprecated-scoped-blocks-imports

Continues #2600 


mklanjsek (approved):
Thanks for doing that, Sam.  When we upgrade to storybook 6.4, we should make decision about `compodoc`, but currently I am leaning towards taking it out.

PR #2647, opened 2021-07-13
Fix form select not sending changes on Submit

### LINKED ISSUE #2623: NewFormFieldSelect cannot be retrieved by request.Payload.StringSlice in octant 0.21.0
I want to upgrade antrea-octant-plugin by using the latest 0.21.0 octant plugin API. However, I met an issue in the latest 0.21.0 octant but the same code can work in octant 0.16.1. 

I used component.NewFormFieldSelect for desTypeCol which is defined here https://github.com/antrea-io/antrea/blob/main/plugins/octant/cmd/antrea-octant-plugin/traceflow.go#L635. Now when I use 
request.Payload.StringSlice(dstTypeCol), it returns the error that shows payload does not contain this field. 

I did not meet this issue in previous 0.16.1 release, so I would like to know if there is any change in recent releases for this part which may cause this issue? Is there anything that my plugin needs to update?
Thanks a lot for your information!

request.Payload.StringSlice() can get the value defined in component.NewFormFieldSelect()

Found some warning logs like below:
W0707 09:51:31.637535   18983 reflector.go:424] /__w/octant/octant/internal/objectstore/dynamic_cache.go:389: watch of *unstructured.Unstructured ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode to metav1.Event") has prevented the request from succeeding

**Environment:**
- Octant version (use `octant version`): 0.21.0
- Kubernetes version (use `kubectl version`):1.19
- OS (macOS 10.15, Windows 10, Ubuntu 19.10 etc):Ubuntu 18.04.3

lenriquez:

Fix problem with select input form sending empty information to the server 

- Fixes #2623

**Release note**:
```
release-note
```

### DIFF
--- a/web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
+++ b/web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
@@ -89,12 +89,14 @@
       <cds-select [style.width.%]="field.config.width">
         <label [for]="field.config.name">{{ field.config.label }}</label>
         <select
-          [formControlName]="field.config.name"
+          [formArrayName]="field.config.name"
           [attr.multiple]="field.config.configuration.multiple"
+          (change)="onSelect($event, field.config.name)"
         >
           <option
             *ngFor="let opt of fieldChoices(field); trackBy: trackByFn"
             [value]="opt.value"
+            [selected]="opt.checked"
           >
             {{ opt.label }}
           </option>
--- a/web/src/app/modules/shared/components/form-view-container/form-view-container.component.spec.ts
+++ b/web/src/app/modules/shared/components/form-view-container/form-view-container.component.spec.ts
@@ -3,6 +3,7 @@ import { ComponentFixture, TestBed, waitForAsync } from '@angular/core/testing';
 import { FormViewContainerComponent } from './form-view-container.component';
 import { Component } from '@angular/core';
 import {
+  FormArray,
   FormBuilder,
   FormGroup,
   FormsModule,
@@ -25,6 +26,8 @@ class TestWrapperComponent {
 describe('FormViewContainerComponent', () => {
   let component: TestWrapperComponent;
   let fixture: ComponentFixture<TestWrapperComponent>;
+  let element: HTMLDivElement;
+  let formHelper;
 
   const formBuilder: FormBuilder = new FormBuilder();
 
@@ -41,6 +44,8 @@ describe('FormViewContainerComponent', () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(TestWrapperComponent);
     component = fixture.componentInstance;
+    element = fixture.nativeElement;
+    formHelper = new FormHelper();
   });
 
   it('should create', () => {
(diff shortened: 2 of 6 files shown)


wwitzel3:
Please add a changelog entry for this since it directly fixes a bug.

lenriquez:
@wwitzel3 Thanks! I forgot

web/src/app/modules/shared/components/form-view-container/form-view-container.component.spec.ts
```diff
@@ -41,6 +43,12 @@ describe('FormViewContainerComponent', () => {
   beforeEach(() => {
     fixture = TestBed.createComponent(TestWrapperComponent);
     component = fixture.componentInstance;
+    element = fixture.nativeElement;
+    formHelper = new FormHelper();
+  });
+
+  it('should create', () => {
```
GuessWhoSamFoo, line 50: This test is duplicated below?

lenriquez: Thanks!

web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
```diff
@@ -89,8 +89,9 @@ <h6 *ngIf="field.config.label" slot="label">{{ field.config.label }}</h6>
       <cds-select [style.width.%]="field.config.width">
         <label [for]="field.config.name">{{ field.config.label }}</label>
         <select
-          [formControlName]="field.config.name"
+          [formArrayName]="field.config.name"
           [attr.multiple]="field.config.configuration.multiple"
+          (change)="onSelect($event, field.config.name)"
         >
           <option
             *ngFor="let opt of fieldChoices(field); trackBy: trackByFn"
```
GuessWhoSamFoo, line 97: Missing `[selected]="opt.checked"` here if we want default selection to work

web/src/app/modules/shared/components/form-view-container/form-view-container.component.spec.ts
```diff
@@ -82,5 +84,38 @@ describe('FormViewContainerComponent', () => {
...
+          },
+        ],
+      };
+      component.formGroup = formHelper.createFromGroup(
+        component.form,
+        formBuilder
+      );
+      fixture.detectChanges();
+
+      expect(element.querySelector('cds-select')).not.toBeNull();
```
GuessWhoSamFoo, line 118: Since choices are added within the test, it would be reasonable to check if the input choices are generated as expected.

web/src/app/modules/shared/components/form-view-container/form-view-container.component.html
lenriquez: Good Catch! [selected]="opt.checked" does not work, I don't think this solution works, I think is better just to use the regular Angular way

lenriquez: Actually it does work but is not reflected on the HTML 
image

GuessWhoSamFoo (approved):

PR #2661, opened 2021-07-16
Upgrade to Angular 12, Clarity 5.4, Typescript 4.5

mklanjsek:
This change brings our code base to latest versions of Angular, Clarity and Typescript.

Details:
- Bumped Angular `11.2.13` -> `12.1.2`,
- Bumped Clarity Core `5.2.0` -> `5.4.2`, 
- Bumped Clarity Angular `5.2.0` -> `12.0.0-next.0` (first release that supports Angular 12),
- Bumped Typescript `4.1.5` -> `4.3.5`,
- Removed functionality that was disabling warnings from storybook (that stopped working a few releases ago and now does not even compile with webpack 5),
- Implemented workarounds for a couple of Clarity issues (will open an issue to remove that code when Clarity  issues are resolved),
- Various small cleanup changes performed by Angular migration tool,


web/angular.json
```diff
@@ -86,7 +91,6 @@
               "outputHashing": "all",
               "sourceMap": false,
               "namedChunks": false,
-              "aot": true,
```
lenriquez, line 89: Just Curious, Is this because in angular 12 AOT is done by default?

mklanjsek: Yes, the migration tool changed that since `aot` is on by default now, see [here](https://blog.ninja-squad.com/2021/05/12/angular-cli-12.0/)

lenriquez: ohh thanks!

mklanjsek:
Netlify deploy is failing because it's using node `v10.24.1` that's not supported by angular 12. Working with @jonasrosland to resolve that.

web/src/app/modules/shared/shared.module.ts
```diff
@@ -92,8 +92,10 @@ import { IconComponent } from './components/presentation/icon/icon.component';
 import { FormViewContainerComponent } from './components/form-view-container/form-view-container.component';
 import { SignpostComponent } from './components/presentation/signpost/signpost.component';
 import { JoinPipe } from './pipes/join/join.pipe';
+import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
```
GuessWhoSamFoo, line 95: This is not needed

GuessWhoSamFoo (commented):
For all instances of `[disabled]="myvar"` in cds-button, replace with `[attr.disabled]="myvar ? true: null"`

web/package.json
```diff
@@ -42,21 +41,21 @@
...
+    "@angular/platform-browser-dynamic": "^12.1.2",
+    "@angular/router": "^12.1.2",
+    "@cds/angular": "^5.4.2",
     "@cds/city": "^1.1.0",
-    "@cds/core": "^5.2.0",
-    "@clr/angular": "^5.2.0",
-    "@clr/icons": "^5.2.0",
-    "@clr/ui": "^5.2.0",
+    "@cds/core": "^5.4.2",
+    "@clr/angular": "^12.0.0-next.0",
```
GuessWhoSamFoo, line 56: It also looks like 12.0.0 just came out

web/src/app/modules/shared/shared.module.ts
mklanjsek: Unfortunately it's needed temporarily as a workaround for [see Clarity issue](https://github.com/vmware/clarity/issues/6080#issuecomment-862742168)

web/package.json
mklanjsek: Bumped both Clarity and Angular since new versions are out for both.

mklanjsek:
> For all instances of `[disabled]="myvar"` in cds-button, replace with `[attr.disabled]="myvar ? true: null"`

Similar as above, Clarity team is recommending using `[disabled]` instead, see same [issue](https://github.com/vmware/clarity/issues/6080#issuecomment-862742168)

mklanjsek:
PTAL

Bumped to versions that were released since this PR was created: Angular to `12.1.3`, Clarity to `5.5.0` and Clarity Angular to `12.0.0`.

Comments related to Clarity button workaround will be addressed through #2662 once [Clarity button issue](https://github.com/vmware/clarity/issues/6080) is fixed.

GuessWhoSamFoo (approved):

GuessWhoSamFoo:
In a separate issue, we should address remaining usage of clr-icon as there are some minor undesired CSS changes as a result of the angular bump (e.g. helper icon on top right)

PR #2675, opened 2021-07-23
Added status column to Pod list

### LINKED ISSUE #201: Show pod status in Workloads / Pods view
Kubectl seems to do some magic to show a "STATUS" of a given pod:

```
NAME                       READY   STATUS              RESTARTS   AGE
busybox-644dbf5f4b-2sjfs   1/1     Terminating         0          42m
busybox-644dbf5f4b-l5lzn   0/1     ContainerCreating   0          31m
```

I was somewhat surprised when I used Octant to look at the same set of pods. This is what Octant shows:

image

Notice how kubectl says the first pod is `Terminating` and the second is `ContainerCreating`, whereas Octant shows `Running` and `Pending`.

I understand that Octant is showing the pod's `Phase` and not the `Status` (I am actually not sure how kubectl computes this), but it was super confusing at first.

I am wondering if we should follow kubectl here and show `Status` instead of `Phase`, or perhaps add a new column that shows `Status`.

ftovaro:
Signed-off-by: ftovaro <f.tovar12@gmail.com>

Adds a new column called Status that reflects the status of a Pod in addition to the Phase that is already on the table.

image

image

image

- Fixes #201 

**Release note**:
```
release-note
```

### DIFF
--- a/internal/printer/pod.go
+++ b/internal/printer/pod.go
@@ -26,8 +26,8 @@ import (
 )
 
 var (
-	podColsWithLabels    = component.NewTableCols("Name", "Labels", "Ready", "Phase", "Restarts", "Node", "Age")
-	podColsWithOutLabels = component.NewTableCols("Name", "Ready", "Phase", "Restarts", "Node", "Age")
+	podColsWithLabels    = component.NewTableCols("Name", "Labels", "Ready", "Phase", "Status", "Restarts", "Node", "Age")
+	podColsWithOutLabels = component.NewTableCols("Name", "Ready", "Phase", "Status", "Restarts", "Node", "Age")
 	podResourceCols      = component.NewTableCols("Container", "Request: Memory", "Request: CPU", "Limit: Memory", "Limit: CPU")
 )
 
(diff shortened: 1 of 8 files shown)


internal/printer/pod.go
```diff
@@ -152,7 +168,7 @@ func createPodSummaryStatus(pod *corev1.Pod) (*component.Summary, error) {
 	sections.AddText("QoS", string(pod.Status.QOSClass))
 
 	if pod.DeletionTimestamp != nil {
-		summary.SetAlert(component.NewAlert(component.AlertStatusError, component.AlertTypeDefault, "Pod is being deleted", false, nil))
+		summary.SetAlert(component.NewAlert(component.AlertStatusError, component.AlertTypeDefault, "Pod is being deleted*****", false, nil))
```
xtreme-vikram-yadav, line 171: Do we need these additional asterisks?

ftovaro: We definitely don't need it, was part of the testing, they are removed now

internal/printer/pod.go
```diff
@@ -70,6 +70,22 @@ func PodListHandler(ctx context.Context, list *corev1.PodList, opts Options) (co
 
 		row["Phase"] = component.NewText(string(pod.Status.Phase))
 
+		if len(pod.Status.ContainerStatuses) > 0 {
+			lastStatus := pod.Status.ContainerStatuses[len(pod.Status.ContainerStatuses)-1]
```
xtreme-vikram-yadav, line 74: Nitpicking here but the logic to determine the status can go inside a private method.

ftovaro: Could be but don't you think that the logic is too simple for that and also a method would have at least 3 lines of code and here we can have it with just one

xtreme-vikram-yadav (approved):

wwitzel3 (approved):

PR #2685, opened 2021-07-26
printer - parse conditions from Unstructured

### LINKED ISSUE #489: Add crd conditions to crd view
Blocked by #566

### LINKED ISSUE #566: In overview summary generate conditions table if conditions are detected
In the overview summary, we have hardcoded conditions tables for certain GVKs. We can simplify this by automatically adding conditions if we detect `.status.conditions`. The win is that any object (including CRDs) that has conditions would automatically get this table.

wwitzel3:

First step in making the Conditions table conditionally available.

**Note for reviewers**
I added a couple extra commits to this PR and it now fixes both issues.

- Fixes #566
- Fixes #489

### DIFF
--- a/internal/printer/conditions.go
+++ b/internal/printer/conditions.go
@@ -1,0 +1,99 @@
+/*
+Copyright (c) 2021 the Octant contributors. All Rights Reserved.
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package printer
+
+import (
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/vmware-tanzu/octant/pkg/view/component"
+)
+
+var (
+	conditionType           = "Type"
+	conditionReason         = "Reason"
+	conditionStatus         = "Status"
+	conditionMessage        = "Message"
+	conditionLastUpdate     = "Last Update"
+	conditionLastTransition = "Last Transition"
+
+	conditionMap = [][]string{
+		{conditionType, "type"},
+		{conditionReason, "reason"},
+		{conditionStatus, "status"},
+		{conditionMessage, "message"},
+		{conditionLastUpdate, "lastUpdateTime"},
+		{conditionLastTransition, "lastTransitionTime"},
+	}
+)
+
+// createConditionsTable returns a component.Table, if conditions exist on the object, and any err.
+// For objects with empty conditions, an empty table with placeholder text is used.
+func createConditionsTable(u *unstructured.Unstructured, sortKey string, customConditionMap [][]string) (*component.Table, bool, error) {
+	if u == nil {
+		return nil, false, errors.New("object is nil")
+	}
+
+	localConditionMap := conditionMap
+	if customConditionMap != nil {
+		localConditionMap = customConditionMap
+	}
+
+	table := component.NewTable("Conditions", "There are no conditions!", nil)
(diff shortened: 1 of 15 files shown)


internal/printer/conditions_test.go
```diff
@@ -0,0 +1,75 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+
+	"github.com/vmware-tanzu/octant/internal/testutil"
+	"github.com/vmware-tanzu/octant/pkg/view/component"
+)
+
+func Test_createConditionsTableErrs(t *testing.T) {
+	// No status found
```
GuessWhoSamFoo, line 18: Any strong opinion on shifting to something like?
```
	cases := []struct {
		name   string
		status unstructured.Unstructured
		err    error
		isOk   bool
		table  *component.Table
	}{
		{
			name:   "no status",
			status: unstructured.Unstructured{Object: map[string]interface{}{"noStatus": nil}},
		},
		{
			name:   "bad status",
			status: unstructured.Unstructured{Object: map[string]interface{}{"status": 1}},
			err:    fmt.Errorf(".status accessor error: 1 is of the type int, expected map[string]interface{}"),
		},
		{
			name:   "no conditions",
			status: unstructured.Unstructured{Object: map[string]interface{}{"status": map[string]interface{}{"noConditions": nil}}},
			table:  component.NewTable("Conditions", "There are no conditions!", nil),
		},
	}

	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			table, ok, err := createConditionsTable(tc.status, "", nil)
			if tc.err == nil {
				require.NoError(t, err)
			} else {
				require.Errorf(t, tc.err, ".status accessor error: 1 is of the type int, expected map[string]interface{}", err)
			}
			require.Equal(t, tc.isOk, ok)
			component.AssertEqual(t, tc.table, table)
		})
	}
```

internal/printer/conditions.go
```diff
@@ -0,0 +1,128 @@
...
+		{conditionStatus, "status"},
+		{conditionMessage, "message"},
+		{conditionLastUpdate, "lastUpdateTime"},
+		{conditionLastTransition, "lastTransitionTime"},
+	}
+)
+
+// createConditionsTable returns a component.Table, if conditions exist on the object, and any err.
+// For objects with empty conditions, an empty table with placeholder text is used.
+func createConditionsTable(u unstructured.Unstructured, sortKey string, customConditionMap [][]string) (*component.Table, bool, error) {
```
GuessWhoSamFoo, line 41: I think this shouldn't return a bool and an exportable `HasConditions(u unstructured.Unstructured)` should be available to check rather than always attempting to make a table

internal/printer/conditions_test.go
wwitzel3: Since you already coded it up, no strong opinion at all. :)

internal/printer/conditions.go
wwitzel3: I agree with essence of this idea, I'll push up a change that I think satisfies this.

wwitzel3:
CRDs now showing conditions
image

wwitzel3:
image

wwitzel3:
CRD with Status but no set conditions
image

wwitzel3:
ping @GuessWhoSamFoo or @lenriquez can I get another pass for changes and or approval

GuessWhoSamFoo (approved):

PR #2692, opened 2021-07-27
Enable go plugins to send events

GuessWhoSamFoo:
Javascript plugins have a `SendEvent` method that is currently not possible with go plugins.

see https://github.com/vmware-tanzu/octant/blob/447171cf3ac3c520c3ad3bbe839f8affc6b61cc3/pkg/plugin/javascript/dashboard_send_event.go

This PR allows plugin authors to do the same via:
`request.DashboardClient.SendEvent(request.Context(),request.ClientState.ClientID(), event.EventTypeMyEvent, action.Payload{})`

cc @liamrathke


wwitzel3 (changes requested):
Lets update SendAlert to wrap SendEvent and we can later talk about removing SendAlert

pkg/plugin/api/server.go
```diff
@@ -312,7 +325,7 @@ func (c *grpcServer) Create(ctx context.Context, in *proto.CreateRequest) (*prot
 	return &proto.CreateResponse{}, nil
 }
 
-// Get gets an object.
+// Delete gets an object.
```
liamrathke, line 328: Super minor nitpick, might want to update this comment to `Delete deletes an object` to avoid potential confusion

pkg/plugin/api/client.go
```diff
@@ -274,3 +276,22 @@ func (c *Client) CreateLink(ctx context.Context, key store.Key) (LinkResponse, e
 		Link: *linkComponent,
 	}, nil
 }
+
+// SendEvent sends an event
+func (c *Client) SendEvent(ctx context.Context, clientID string, eventName event.EventType, payload action.Payload) error {
```
liamrathke, line 281: Agree with @wwitzel3 that we should modify `SendAlert()` to use `SendEvent()`.

pkg/plugin/api/server.go
```diff
@@ -379,14 +380,18 @@ func (c *grpcServer) ForceFrontendUpdate(ctx context.Context, _ *proto.Empty) (*
 	return &proto.Empty{}, nil
 }
 
-// SendAlert sends an alert
+// SendAlert sends an alert. This method is deprecated and will be removed in a future release
```
liamrathke, line 383: Just to be clear, we'd just need to deprecate it from the server side, right? As in, we could modify `SendAlert()` in `client.go` to call `client.SendEvent()`, so that plugins sending alerts won't need to be rewritten. I feel like `SendAlert()` is a good utility to have client-side, since it's widely usable. Were it removed, any plugins would need to re-implement the same code, and it would look almost identical to what we have already anyways. 

https://github.com/vmware-tanzu/octant/blob/f2b531ca649e23e711f9118019ea185e8412626f/pkg/plugin/api/client.go#L242-L255

GuessWhoSamFoo: The idea is to remove it entirely because it is redundant (and we can do so relatively safely as it is pre 1.0). Otherwise one could argue one should have similar methods for each event type.

The tradeoff is abstracting away knowledge of the required `action.Payload` keys which is something that could be addressed by creating payload helpers for various events.

liamrathke: Makes sense, some sort of action/event helpers would be super useful for plugins

liamrathke (approved):

wwitzel3 (approved):

PR #2695, opened 2021-07-29
Show container image manifest for pods

### LINKED ISSUE #156: Show image manifest for containers
Show image manifest for containers, so users can know exactly which
image they are running. This will be helpful if you are using 
images with `latest` tag.

mklanjsek:
Added two new fields to Pod Container section: `Image Manifest` and `Image Configuration` as shown here in partially expanded state:

Screen Shot 2021-07-29 at 10 11 19 AM

The output for these two fields is generated by `containers\image` go library, same library used for `containers\skopeo` tool. Image Manifest corresponds to output of following skopeo command:
```skopeo inspect --raw docker://nginx:1.16.1```

Image configuration is showing image configuration based on current OS, similar to output of: 
```skopeo inspect --config --override-os linux docker://alpine:3.14.0```
 
Note that for Mac library only supports `linux` OS setting, so it has to be set manually to override the OS value.

Tested only on Mac, need help testing it on other operating systems. 

- Fixes #156

### CODE REVIEW

**Comment by karayim** [comment]
The green looks much better in the dark mode. Thanks for taking a look at it!

**Comment by wwitzel3** [comment]
I'm worried about the dependencies here, specifically with building for Windows and in our CI pipeline. I don't have any answers. We can probably get this building under ubuntu by adding a call to `sudo apt-get install libbtrfs` or something on those lines.

Unsure about Mac and Windows.

**Comment by mklanjsek** [comment]
After adding additional tags to all go build commands (`-tags exclude_graphdriver_devicemapper exclude_graphdriver_btrfs containers_image_openpgp`) everything is working except the verification of generated code. Looks like the `go generate` is behaving differently in terms how tags are passed in.

Additionally, windows tests were failing due to inability to locate windows alpine and nginx images. Temporarily fixed that by always overriding OS to `linux`. Not sure that's the right thing to do, we need to discuss that and do more testing on Windows.

**Comment by wwitzel3** [comment]
If we want to use thing library, we are going to have to do some work upstream to support Windows. As mentioned in the issue, currently Windows just has stubs for the methods.

https://github.com/containers/storage/issues/503#issuecomment-666542384

**Comment by wwitzel3** [comment]
I think at this point we may want to see if there are any library alternatives that get us what we need across platforms or look in to how much of a lift it is to get this working correctly on Windows.

**Comment by wwitzel3** [comment]
We many also be able to abstract the docker manifest code directly and leverage it as a library.

https://github.com/docker/cli/blob/a32cd16160f1b41c1c4ae7bee4dac929d1484e59/cli/command/manifest/inspect.go#L94

**Comment by wwitzel3** [comment]
Did some more investigating after @mklanjsek pointed out the issue in question, that code path wasn't being called with his use of the the library and so this should all be working under Windows.

Found that configuring mingw64 and the appropriate deps using MSYS resulting in a successful compilation under Windows.

I think for our windows build, we can add https://github.com/marketplace/actions/install-mingw

We will want to configure the Windows job with on of the extra actions below
- https://github.com/msys2/setup-msys2
- https://github.com/marketplace/actions/install-mingw

**Comment by wwitzel3** [comment]
For the verify generated code, add `- run: sudo apt-get update && sudo apt install -y libgpgme-dev libassuan-dev libdevmapper-dev btrfs-progs-devel` as a step to https://github.com/vmware-tanzu/octant/blob/5163d0f17f41f99b8f8a72de936224998228aabd/.github/workflows/verify-generated.yaml#L16

Should fix that up.

File: internal/printer/container.go [hunk 1]
```diff
@@ -122,6 +128,14 @@ func (cc *ContainerConfiguration) Create() (*component.Summary, error) {
 		sections.AddText("Image ID", containerStatus.ImageID)
 	}
 
+	manifest, configuration, err := ManifestManager.GetImageManifest(cc.context, hostOS, c.Image)
+	if err == nil {
+		sections.Add("Image Manifest", component.NewJSONEditor(manifest, true))
+		sections.Add("Image Configuration", component.NewJSONEditor(configuration, true))
+	} else {
+		sections.Add("Image Manifest", component.NewText(fmt.Sprintf("Unable to load image manifest %s", err)))
```
xtreme-vikram-yadav, line 136: The error shown in the UI is hard to read when all the errors from the error chain are concatenated together.
An easy way to reproduce this is to kill the internet before caching and see the error in the pod summary.
It shows something like:
```Unable to load image manifest error creating image source for image docker://nginx:1.14.2: pinging container registry registry-1.docker.io: Get "https://registry-1.docker.io/v2/": dial tcp: lookup registry-1.docker.io: no such host```.
The error from https://github.com/vmware-tanzu/octant/pull/2695/files#diff-85fedc5e144e89139dc0296c0fb2574d2576d493b8bad648bd79c31d32dbd73bR63 and others in the manifest.go and any subsequent errors can be formatted a little better with some delimiters.

xtreme-vikram-yadav (changes requested):

internal/printer/container.go
xtreme-vikram-yadav: This change is not a huge blocker and can be made afterwards.

PR #2774, opened 2021-08-18
Add protoc-gen-go-grpc

### LINKED ISSUE #2691: Remove deprecated usage of pb.Timestamp
From github.com/golang/protobuf/ptypes,
```
// Timestamp converts a timestamppb.Timestamp to a time.Time.
// It returns an error if the argument is invalid.
//
// Unlike most Go functions, if Timestamp returns an error, the first return
// value is not the zero time.Time. Instead, it is the value obtained from the
// time.Unix function when passed the contents of the Timestamp, in the UTC
// locale. This may or may not be a meaningful time; many invalid Timestamps
// do map to valid time.Times.
//
// A nil Timestamp returns an error. The first return value in that case is
// undefined.
//
// Deprecated: Call the ts.AsTime and ts.CheckValid methods instead.
```

Remove usage before breaking changes.

**Environment:**

- Octant version (use `octant version`): 0.22

GuessWhoSamFoo:
Per Readme in https://github.com/golang/protobuf, https://pkg.go.dev/mod/google.golang.org/protobuf is the preferred module which splits the protobuf compiler (protoc) and the runtime modules.

This update requires adding `mustEmbedUnimplemented__` for each registered service. See https://github.com/grpc/grpc-go/issues/3669 for background on this change.

- Fixes #2691


xtreme-vikram-yadav (approved):

GuessWhoSamFoo:
I'm going to hold this until all other existing work with protobuf related changes are merged this sprint

PR #2818, opened 2021-09-01
plugins - automatically reload/remove Go plugins

### LINKED ISSUE #178: Installing a plugin requires restart of main octant process.
I installed the sample plugin and it did not show up until I restarted.

wwitzel3:

Currently Go plugins do not automatically reload or remove themselves based on changes happening to the binaries in the plugin folder and a full restart of Octant is required to view plugin changes. We want to enable the same behavior we have for the JavaScript plugins for the Go plugins.

- If a binary is removed, this is treated as an unload operation.
- If a binary is created, this is treated as a load operation.
- If a binary is renamed, this is treated as an unload (old name), load (new name) operation.
- If a binary is replaced, this is treated as an unload, load operation.

- Fixes #178

### DIFF
--- a/pkg/plugin/manager.go
+++ b/pkg/plugin/manager.go
@@ -640,7 +640,7 @@ func (m *Manager) Start(ctx context.Context) error {
 	return nil
 }
 
-// goPluginPingPong will attempt to restart a Go plugin process if the Ping method for a client returns a non-nill error.
+// goPluginPingPong will attempt to restart a Go plugin process if the Ping method for a client returns a non-nil error.
 func (m *Manager) goPluginPingPong(ctx context.Context) {
 	logger := log.From(ctx)
 
@@ -668,6 +668,8 @@ func (m *Manager) goPluginPingPong(ctx context.Context) {
 						logger.WithErr(err).Errorf("unable to find command for plugin")
 						continue
 					}
+
+					m.Unload(ctx, cmd)
 
 					c := PluginConfig{
 						Name: clientName,
--- a/pkg/plugin/manager.go
+++ b/pkg/plugin/manager.go
@@ -82,10 +82,13 @@ type Client interface {
 type ManagerStore interface {
 	Store(name string, client Client, metadata *Metadata, cmd string) error
 	StoreJS(name string, jspc JSPlugin) error
+	Get(name string) (Client, bool)
 	GetJS(name string) (JSPlugin, bool)
+	Remove(name string)
 	RemoveJS(name string)
 	NamesJS() []string
 	GetMetadata(name string) (*Metadata, error)
+	GetModuleService(name string) (ModuleService, error)
 	GetService(name string) (Service, error)
 	GetCommand(name string) (string, error)
 	Clients() map[string]Client
@@ -130,6 +133,11 @@ func (s *DefaultStore) StoreJS(name string, plugin JSPlugin) error {
 	return nil
 }
 
+func (s *DefaultStore) Get(name string) (Client, bool) {
+	client, ok := s.clients[name]
+	return client, ok
+}
+
 func (s *DefaultStore) GetJS(name string) (JSPlugin, bool) {
 	voidStar, ok := s.jsPlugins.Load(name)
 	if !ok {
@@ -137,6 +145,12 @@ func (s *DefaultStore) GetJS(name string) (JSPlugin, bool) {
 	}
 	jspc, ok := voidStar.(JSPlugin)
 	return jspc, ok
+}
+
+func (s *DefaultStore) Remove(name string) {
+	delete(s.commands, name)
+	delete(s.clients, name)
+	delete(s.metadata, name)
 }
 
 func (s *DefaultStore) RemoveJS(name string) {
@@ -156,8 +170,8 @@ func (s *DefaultStore) Store(name string, client Client, metadata *Metadata, cmd
 	return nil
 }
 
-// GetService gets the service for a plugin.
-func (s *DefaultStore) GetService(name string) (Service, error) {
+// GetModuleService gets the moduleService for a plugin
+func (s *DefaultStore) GetModuleService(name string) (ModuleService, error) {
 	client, ok := s.clients[name]
 	if !ok {
 		return nil, errors.Errorf("plugin %q doesn't have a client", name)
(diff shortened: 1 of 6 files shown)


pkg/plugin/manager.go
```diff
@@ -82,10 +82,13 @@ type Client interface {
 type ManagerStore interface {
 	Store(name string, client Client, metadata *Metadata, cmd string) error
 	StoreJS(name string, jspc JSPlugin) error
+	Get(name string) (Client, bool)
 	GetJS(name string) (JSPlugin, bool)
+	Remove(name string)
```
GuessWhoSamFoo, line 87: If a plugin was delete while the user was looking at a view, what should happen?

Currently https://github.com/vmware-tanzu/octant/blob/master/internal/api/content_manager.go#L191 returns the error each poll, but we might want to redirect back to a 404-Like page

pkg/plugin/manager.go
```diff
@@ -554,12 +634,14 @@ func (m *Manager) Start(ctx context.Context) error {
...
 
-	go m.watchPlugins(ctx)
+	go m.watchPluginFiles(ctx)
+	go m.goPluginPingPong(ctx)
 
 	return nil
 }
 
-func (m *Manager) watchPlugins(ctx context.Context) {
+// goPluginPingPong will attempt to restart a Go plugin process if the Ping method for a client returns a non-nill error.
```
GuessWhoSamFoo, line 643: nit: nil

GuessWhoSamFoo:
Not necessarily something to be addressed in this PR: whenever a plugin panics, the content manager stops returning any data.

wwitzel3:
> Not necessarily something to be addressed in this PR: whenever a plugin panics, the content manager stops returning any data.

Yeah, I think that would be out of scope for this. That is currently the case before this change. I'll create an issue for it.

wwitzel3:
> I faced some confusion while testing this by updating the plugin by repeatedly running `go run build.go install-test-plugin`. This only happens when I use the `build.go` because I am used to it. I see a log entry `2021-09-01T19:36:16.921-0400 INFO plugin/manager.go:448 removing plugin: C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe 2021-09-01T19:36:17.991-0400 INFO plugin/manager.go:424 reloading: Go plugin: C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe~`.
> But the original plugin binary was not deleted and it cannot be delete even as an admin because it is still running and interacting with octant. So, windows creates another binary by appending ~ (octant-sample-plugin.exe~). This causes octant to register the same plugin twice because the name of the plugin is technically different.
> duplicate-plugin
> 
> Octant fails again if we try to install the plugins with more updates but since windows cannot create more file with a same name (I guess it's last attempt to resolve file name conflicts is to append ~) and it fails to replace the existing binary(octant-sample-plugin.exe~) and throws the following error:
> 
> ```
> 2021/09/01 19:41:10 Running: C:\Program Files\Go\bin\go.exe build -o C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe github.com/vmware-tanzu/octant/cmd/octant-sample-plugin
> go build github.com/vmware-tanzu/octant/cmd/octant-sample-plugin: copying C:\Users\vikram\AppData\Local\Temp\go-build3339358845\b001\exe\a.out.exe: open C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe: The process cannot access the file because it is being used by another process.
> 2021/09/01 19:41:13 exit status 1
> exit status 1
> ```
> 
> All of this makes sense and this problem can be avoided by creating a new binary with a different name. May be it's worth documenting with this feature in case users run the same script to create a new binary with a same name. We should also update our build script to make `install-test-plugin` a bit more convenient. The build script update is a low priority issue and can be scheduled for later.

So I've been thinking about this and I'm not exactly sure what we can do as part of this set of changes. Windows filesystem provides some unique challenges with respect to processes in use and being able to overwrite files.

> I faced some confusion while testing this by updating the plugin by repeatedly running `go run build.go install-test-plugin`. This only happens when I use the `build.go` because I am used to it. I see a log entry `2021-09-01T19:36:16.921-0400 INFO plugin/manager.go:448 removing plugin: C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe 2021-09-01T19:36:17.991-0400 INFO plugin/manager.go:424 reloading: Go plugin: C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe~`.
> But the original plugin binary was not deleted and it cannot be delete even as an admin because it is still running and interacting with octant. So, windows creates another binary by appending ~ (octant-sample-plugin.exe~). This causes octant to register the same plugin twice because the name of the plugin is technically different.
> duplicate-plugin
> 
> Octant fails again if we try to install the plugins with more updates but since windows cannot create more file with a same name (I guess it's last attempt to resolve file name conflicts is to append ~) and it fails to replace the existing binary(octant-sample-plugin.exe~) and throws the following error:
> 
> ```
> 2021/09/01 19:41:10 Running: C:\Program Files\Go\bin\go.exe build -o C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe github.com/vmware-tanzu/octant/cmd/octant-sample-plugin
> go build github.com/vmware-tanzu/octant/cmd/octant-sample-plugin: copying C:\Users\vikram\AppData\Local\Temp\go-build3339358845\b001\exe\a.out.exe: open C:\Users\vikram\AppData\Local\octant\plugins\octant-sample-plugin.exe: The process cannot access the file because it is being used by another process.
> 2021/09/01 19:41:13 exit status 1
> exit status 1
> ```
> 
> All of this makes sense and this problem can be avoided by creating a new binary with a different name. May be it's worth documenting with this feature in case users run the same script to create a new binary with a same name. We should also update our build script to make `install-test-plugin` a bit more convenient. The build script update is a low priority issue and can be scheduled for later.

I think the solution to this is to create a Windows specific loader that makes a copy of the binary before executing it. Our process control would be around the copy of the binary, but our change detection would be around the original file. I think that this is out of scope for this work specifically.

I will create a new issue.

GuessWhoSamFoo (approved):
The plugin path redirect on delete has been addressed.

Remaining issues are windows specific or already existed so they are out of scope for 0.24

PR #2841, opened 2021-09-08
Hide manifests behind button toggle

GuessWhoSamFoo:
- Creates a handler for a request manifest action
- Moves manifest generation out of printer
- Skips the test in a CI environment as `types.ImageSource` introduces more complexity than desired
- Closes `ImageSource` after usage

https://kubernetes.slack.com/archives/CM37M9FCG/p1630591265028400


internal/printer/container_test.go
```diff
@@ -206,11 +209,7 @@ func Test_ContainerConfiguration(t *testing.T) {
...
+					Content: component.NewButton("Show Manifest", action.CreatePayload(octant.ActionGetManifest, map[string]interface{}{"host": hostOS, "image": validContainer.Image})),
```
xtreme-vikram-yadav, line 212: Since image info is fetched and shown on demand, we can add a test for when that info already exists in cache.

xtreme-vikram-yadav (changes requested):
Looks great overall!!

xtreme-vikram-yadav (approved):

